PR #2, opened 2016-06-28
build: make libyami-utils works on android

YuJiankang:
make libyami-utils works on android. verified with androidplayer, v4l2decode and v4l2encode


YuJiankang:
It's only for android, As there is no requirement for android support from customer currently, it can be dropped. I will improve the patch and submit it again if we get the final decision of Android related things. thanks a lot.

PR #4, opened 2016-07-11
Enforce building warnings as errors

chivakker:
Warnings are treated as errors with this patch.  It follows libyami way and corrects warnings found so far. If more warnings are seen  please report back.

### DIFF
--- a/tests/decodeinputavformat.cpp
+++ b/tests/decodeinputavformat.cpp
@@ -38,11 +38,15 @@ bool DecodeInputAvFormat::initInput(const char* fileName)
         goto error;
     uint32_t i;
     for (i = 0; i < m_format->nb_streams; ++i) {
-        AVCodecContext* ctx = m_format->streams[i]->codec;
+#if LIBAVCODEC_VERSION_INT >= AV_VERSION_INT(57, 33, 100)
+        AVCodecParameters* codec = m_format->streams[i]->codecpar;
+#else
+        AVCodecContext* codec = m_format->streams[i]->codec;
+#endif
         //VP9: width and height of the IVF header,VP8: width and height is zero
-        if (AVMEDIA_TYPE_VIDEO == ctx->codec_type) {
-            width = ctx->width;
-            height = ctx->height;
+        if (AVMEDIA_TYPE_VIDEO == codec->codec_type) {
+            width = codec->width;
+            height = codec->height;
             break;
         }
     }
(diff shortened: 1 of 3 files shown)


tests/decodeoutput.cpp
```diff
@@ -23,7 +23,12 @@
 #include "common/VaapiUtils.h"
 
 #if __ENABLE_MD5__
+// including bsd/md5.h produces a warning with __bounded__ attribute,
+// libyami-utils chooses to ignore this particular warning
+#pragma GCC diagnostic ignored "-Wattributes"
+#pragma GCC diagnostic push
```
uartie: I think the "push" should be before the "ignored"?

chivakker: yeap, you're right, I'll push a new version

tests/decodeinputavformat.cpp
```diff
@@ -38,11 +38,11 @@ bool DecodeInputAvFormat::initInput(const char* fileName)
         goto error;
     uint32_t i;
     for (i = 0; i < m_format->nb_streams; ++i) {
-        AVCodecContext* ctx = m_format->streams[i]->codec;
```
xuguangxin: Could we guard this with some ifdef like this. So we do not force them use any version of ffmpeg.
```
static const MimeEntry MimeEntrys[] = {
    AV_CODEC_ID_MPEG2VIDEO, YAMI_MIME_MPEG2,
    AV_CODEC_ID_VP8, YAMI_MIME_VP8,

#if LIBAVCODEC_VERSION_INT > AV_VERSION_INT(54, 40, 0)
    AV_CODEC_ID_VP9, YAMI_MIME_VP9,
#endif

    AV_CODEC_ID_H264, YAMI_MIME_H264,
    ....
};
```

xuguangxin: it should be simple like this
```
#if LIBAVCODEC_VERSION_INT > AV_VERSION_INT(aa, bb, cc)
    AVCodecParameters* ctx = m_format->streams[i]->codecpar;
#else
   AVCodecContext* ctx = m_format->streams[i]->codec;
#endif
```

tests/decodeinputavformat.cpp
```diff
@@ -52,18 +52,18 @@ bool DecodeInputAvFormat::initInput(const char* fileName)
...
-            if (ctx->coded_width > width)
-                width = ctx->coded_width;
-            if (ctx->coded_height > height)
-                height = ctx->coded_height;
-            setResolution(width, height);
+            if (codecpar->width > width)
+                width = codecpar->width;
+            if (codecpar->height > height)
+                height = codecpar->height;
+            //setResolution(width, height);
```
xuguangxin: why comment out this? VppInputDecode::config need this

tests/decodeinputavformat.cpp
```diff
@@ -52,18 +52,18 @@ bool DecodeInputAvFormat::initInput(const char* fileName)
         goto error;
     for (i = 0; i < m_format->nb_streams; i++)
     {
-        AVCodecContext* ctx = m_format->streams[i]->codec;
```
xuguangxin: Please guard code like previous one.

tests/decodeinputavformat.cpp
chivakker: Yeah this is a mistake.  I'll fix it and resubmit.

tests/decodeinputavformat.cpp
chivakker: IMO, this requires more discussion as which libavcodec version we will support. Once we agree on that I'll change this patch.

chivakker: After realizing that the deprecated parameter was introduced on latest libavcodec version, I think it is good to support at least current release version and one before.

xuguangxin: ffmpeg's api will change rapidly. We cannot follow very version. 
Other possible solution is make ffmpeg as a submodule of this project. We only update the submodule when we need new features.

xuguangxin:
+1

xuguangxin:
merged as 336520bce9dfa81493672b4e2bc66d49b736dcde

PR #6, opened 2016-07-15
encoder_h264:add parameter Cabac Dct8x8 and DeblockFilter

wudping:
the following patches:
0001-encoder_h264-add-parameter-Cabac-Dct8x8-and-DeblockF.patch
0002-encoder_h264-fix-the-bug-of-AVC-deblock.patch 
are the related files in libyami, apache branch.

### DIFF
--- a/tests/vppoutputencode.cpp
+++ b/tests/vppoutputencode.cpp
@@ -17,6 +17,7 @@
 #include "config.h"
 #endif
 #include "vppoutputencode.h"
+#include "YamiVersion.h"
 
 EncodeParams::EncodeParams()
     : rcMode(RATE_CONTROL_CQP)
@@ -28,6 +29,11 @@ EncodeParams::EncodeParams()
     , numRefFrames(1)
     , idrInterval(0)
     , codec("AVC")
+    , enableCabac(true)
+    , enableDct8x8(false)
+    , enableDeblockFilter(true)
+    , deblockAlphaOffsetDiv2(2)
+    , deblockBetaOffsetDiv2(2)
 {
     /*nothing to do*/
 }
@@ -70,7 +76,7 @@ void VppOutputEncode::initOuputBuffer()
 static void setEncodeParam(const SharedPtr<IVideoEncoder>& encoder,
                            int width, int height, const EncodeParams* encParam)
 {
-       //configure encoding parameters
+    //configure encoding parameters
     VideoParamsCommon encVideoParams;
     encVideoParams.size = sizeof(VideoParamsCommon);
     encoder->getParameters(VideoParamsTypeCommon, &encVideoParams);
@@ -88,7 +94,6 @@ static void setEncodeParam(const SharedPtr<IVideoEncoder>& encoder,
     encVideoParams.rcParams.initQP = encParam->initQp;
     encVideoParams.rcMode = encParam->rcMode;
     encVideoParams.numRefFrames = encParam->numRefFrames;
-
     encVideoParams.size = sizeof(VideoParamsCommon);
     encoder->setParameters(VideoParamsTypeCommon, &encVideoParams);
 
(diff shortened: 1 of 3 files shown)


lizhong1008:
H264 encoding patch should not be involved here, it should be sent as libyami pull request.

tests/vppoutputencode.cpp
```diff
@@ -98,6 +102,11 @@ static void setEncodeParam(const SharedPtr<IVideoEncoder>& encoder,
     encoder->getParameters(VideoParamsTypeAVC, &encVideoParamsAVC);
     encVideoParamsAVC.idrInterval = encParam->idrInterval;
     encVideoParamsAVC.size = sizeof(VideoParamsAVC);
+    encVideoParamsAVC.enableCabac = encParam->enableCabac;
```
xuguangxin: build will break if you use old yami core library. better
1. increase api number in yami core library.
2. guard this code with 
```
#if YAMI_CHECK_API_VERSION(0, 2, 1)
......
#endif
```

0001-yamitranscode-add-Cabac-Dct8x8-and-DeblockFilter.patch
```diff
@@ -0,0 +1,239 @@
+From 6d95d5a8bb3f11b82420682343a0c766b54ef15e Mon Sep 17 00:00:00 2001
```
xuguangxin: ... why need this patch.

tests/vppoutputencode.cpp
```diff
@@ -98,7 +103,18 @@ static void setEncodeParam(const SharedPtr<IVideoEncoder>& encoder,
     encoder->getParameters(VideoParamsTypeAVC, &encVideoParamsAVC);
     encVideoParamsAVC.idrInterval = encParam->idrInterval;
     encVideoParamsAVC.size = sizeof(VideoParamsAVC);
+#if YAMI_CHECK_API_VERSION(0, 2, 1)
+    encVideoParamsAVC.enableCabac = encParam->enableCabac;
+    encVideoParamsAVC.enableDct8x8 = encParam->enableDct8x8;
+    encVideoParamsAVC.enableDeblockFilter = encParam->enableDeblockFilter;
+    encVideoParamsAVC.deblockAlphaOffset = encParam->deblockAlphaOffset;
+    encVideoParamsAVC.deblockBetaOffset = encParam->deblockBetaOffset;
     encoder->setParameters(VideoParamsTypeAVC, &encVideoParamsAVC);
```
xuguangxin: should put this after #endif

wudping: Thanks Guangxin.

tests/vppoutputencode.cpp
```diff
@@ -98,6 +103,17 @@ static void setEncodeParam(const SharedPtr<IVideoEncoder>& encoder,
     encoder->getParameters(VideoParamsTypeAVC, &encVideoParamsAVC);
     encVideoParamsAVC.idrInterval = encParam->idrInterval;
     encVideoParamsAVC.size = sizeof(VideoParamsAVC);
+#if YAMI_CHECK_API_VERSION(0, 2, 1)
```
lizhong1008: Could we add version checking at congfigure file? Else we always need it when interface updated, it make our code looks ugly.

tests/yamitranscode.cpp
```diff
@@ -49,6 +49,11 @@ static void print_help(const char* app)
     printf("   --intraperiod <Intra frame period(default 30)> optional\n");
     printf("   --refnum <number of referece frames(default 1)> optional\n");
     printf("   --idrinterval <AVC/HEVC IDR frame interval(default 0)> optional\n");
+    printf("   --enablecabac <AVC is to use CABAC or not in Main Profile (default true)> optional\n");
+    printf("   --enabledct8x8 <AVC is to use DCT8x8 transform or not in High Profile (default false)> optional\n");
+    printf("   --enabledeblockfilter <AVC is to use Deblock filter or not (default true)> optional\n");
+    printf("   --deblockalphadiv2 <AVC Alpha offset of debloking filter divided 2 (default 0)> optional\n");
+    printf("   --deblockbetadiv2 <AVC Beta offset of debloking filter divided 2 (default 0)> optional\n");
```
lizhong1008: The default value is 2, not 0.

tests/vppoutputencode.cpp
xuguangxin: It is a tradeoff, check it in configure will introduce huge efforts to bisect.
Suppose I find an issue, I need check which commit bring this in.  Use this approach we do not need change yami-utils. We can just bisect yami core library. However, if we use configure approach. We need find every workable pair of yami core library and yami-utils. It is a nightmare for bug tracker.

lizhong1008: Yeah, seems it different from libva. Both VAAPI interface and test case are put in libva project. It is a tough thing, let's keep it right now.

lizhong1008:
+1

lizhong1008:
Merged as 2c03fe5da3d4c47f084c7ea218b329a9ceab126b.  Thanks, Dongping.

PR #8, opened 2016-07-21
We should not abort all deocde process if it's only non-fatal error returned.

xuguangxin:
Bitstream have error bits some time. We can skip decode the frame. But we'd better do not abort entire deocde process. So we ignore not-fatal error.


xuguangxin:
Hi @lizhong1008 
Help to review this too.
thanks

lizhong1008:
+1

xuguangxin:
merged as 4e36e1ca119f97e76c5271895483b4fe651945d1

PR #11, opened 2016-08-02
transcode: set ipPeriod to poolsize #461

wudping:
replace const poolsize 5 with ipPeriod in yamitranscode.

### DIFF
--- a/tests/yamitranscode.cpp
+++ b/tests/yamitranscode.cpp
@@ -244,11 +244,11 @@ SharedPtr<VppOutput> createOutput(TranscodeParams& para, const SharedPtr<VADispl
     return output;
 }
 
-SharedPtr<FrameAllocator> createAllocator(const SharedPtr<VppOutput>& output, const SharedPtr<VADisplay>& display)
+SharedPtr<FrameAllocator> createAllocator(const SharedPtr<VppOutput>& output, const SharedPtr<VADisplay>& display, int32_t extraSize)
 {
     uint32_t fourcc;
     int width, height;
-    SharedPtr<FrameAllocator> allocator(new PooledFrameAllocator(display, 5));
+    SharedPtr<FrameAllocator> allocator(new PooledFrameAllocator(display, std::max(extraSize, 5)));
     if (!output->getFormat(fourcc, width, height)
         || !allocator->setFormat(fourcc, width,height)) {
         allocator.reset();
@@ -280,7 +280,7 @@ public:
             ERROR("create input or output failed");
             return false;
         }
-        m_allocator = createAllocator(m_output, m_display);
+        m_allocator = createAllocator(m_output, m_display, m_cmdParam.m_encParams.ipPeriod);
         return m_allocator;
     }
 


tests/yamitranscode.cpp
```diff
@@ -244,11 +244,15 @@ SharedPtr<VppOutput> createOutput(TranscodeParams& para, const SharedPtr<VADispl
     return output;
 }
 
-SharedPtr<FrameAllocator> createAllocator(const SharedPtr<VppOutput>& output, const SharedPtr<VADisplay>& display)
+SharedPtr<FrameAllocator> createAllocator(const SharedPtr<VppOutput>& output, const SharedPtr<VADisplay>& display, int32_t poolsize)
```
xuguangxin: better name it as extraSize

tests/yamitranscode.cpp
```diff
@@ -244,11 +244,15 @@ SharedPtr<VppOutput> createOutput(TranscodeParams& para, const SharedPtr<VADispl
...
 {
     uint32_t fourcc;
     int width, height;
-    SharedPtr<FrameAllocator> allocator(new PooledFrameAllocator(display, 5));
+
+    if(poolsize < 5){
```
xuguangxin: Better change to this:
`SharedPtr<FrameAllocator> allocator(new PooledFrameAllocator(display, std::max(extraSize, 5)));`

wudping: okay, thanks, Guangxin.

tests/yamitranscode.cpp
wudping: okay, thanks, Guangxin.

xuguangxin:
could we have some descriptive information in  commit title and detial.
Like fixed issue id. Why we need do this fix...

xuguangxin:
+1

xuguangxin:
@mypopydev 
This will fixes https://github.com/01org/libyami/issues/461. Could you help check it?
thanks

tests/yamitranscode.cpp
```diff
@@ -244,11 +244,11 @@ SharedPtr<VppOutput> createOutput(TranscodeParams& para, const SharedPtr<VADispl
     return output;
 }
 
-SharedPtr<FrameAllocator> createAllocator(const SharedPtr<VppOutput>& output, const SharedPtr<VADisplay>& display)
+SharedPtr<FrameAllocator> createAllocator(const SharedPtr<VppOutput>& output, const SharedPtr<VADisplay>& display, int32_t extraSize)
 {
     uint32_t fourcc;
     int width, height;
-    SharedPtr<FrameAllocator> allocator(new PooledFrameAllocator(display, 5));
+    SharedPtr<FrameAllocator> allocator(new PooledFrameAllocator(display, std::max(extraSize, 5)));
```
chivakker: Why is this not using the SurfacePoolAllocator?  Each codec can control how many surfaces are allocated, so why not use it?

xuguangxin: Two types of surface included in encoder. One for input yuv, one for reconstruct yuv. Encoder can decide the size of reconstruct yuv list. However, for input yuv, we handle it in two ways:
1.  If you use `virtual YamiStatus encode(VideoFrameRawData* frame) = 0;`, we create a new surface and copy content for you.
2. If you use `virtual YamiStatus encode(const SharedPtr<VideoFrame>& frame) = 0;`, we will simply put the frame to reorder list. This mode is for zero copy encode, you can think the frame is from camera or egl texture. 
This patch fix issue for scenario 2, suppose you want set the encode sequence as IBBBBP, you need allocate at least 6 frames for input surface list, because we will cache all B frames until we have P frame.

seanvk: Thanks, for the explanation, lgtm +1

xuguangxin:
merged as 99ff8c2ad710f4e697f7a338d3e446b6c2a95bf1

tests/yamitranscode.cpp
xuguangxin: welcome;)

PR #13, opened 2016-08-04
configure.ac: check libyami version

xuguangxin:
To avoid add nasty version check in source code like this
https://github.com/01org/libyami-utils/blob/master/tests/vpp.cpp#L212.
We’d better check min libyami version number in configure.ac.
This will break automatic “git bisect”.However, it will keep our code more tidy.


chivakker:
lgtm :+1:

PR #15, opened 2016-08-10
Vp9enc enabled on yamiencode

chivakker:
These changes add modifications to have yamiencode handling vp9 encoding.  Since this is also modifying the vp8 path the patches are submitted for consideration.  VP9 encoder has to be enabled on libyami to operate but VP8 encoder should behave as usual.

### DIFF
--- a/tests/encodehelp.h
+++ b/tests/encodehelp.h
@@ -59,8 +59,8 @@ static void print_help(const char* app)
     printf("   -o <coded file> optional\n");
     printf("   -b <bitrate: kbps> optional\n");
     printf("   -f <frame rate> optional\n");
-    printf("   -c <codec: AVC|VP8|JPEG> Note: not support now\n");
-    printf("   -s <fourcc: NV12|IYUV|YV12> Note: not support now\n");
+    printf("   -c <codec: AVC|VP8|VP9|JPEG>\n");
+    printf("   -s <fourcc: NV12|IYUV|YV12>\n");
     printf("   -N <number of frames to encode(camera default 50), useful for camera>\n");
     printf("   --qp <initial qp> optional\n");
     printf("   --rcmode <CBR|CQP> optional\n");
--- a/tests/encodeinput.cpp
+++ b/tests/encodeinput.cpp
@@ -167,43 +167,49 @@ EncodeOutput::~EncodeOutput()
         fclose(m_fp);
 }
 
-EncodeOutput* EncodeOutput::create(const char* outputFileName, int width , int height)
+EncodeOutput* EncodeOutput::create(const char* outputFileName, int width,
+                                   int height, const char* codecName)
 {
-    EncodeOutput * output = NULL;
-    if(outputFileName==NULL)
+    EncodeOutput* output = NULL;
+    if (outputFileName == NULL)
         return NULL;
-    const char *ext = strrchr(outputFileName,'.');
-    if(ext==NULL)
+    const char* ext = strrchr(outputFileName, '.');
+    std::string codec(""); // needed for strcasecmp to work
+    if (!ext && !codecName)
         return NULL;
-    ext++;//h264;264;jsv;avc;26l;jvt;ivf
-    if(strcasecmp(ext,"h264")==0 ||
-        strcasecmp(ext,"264")==0 ||
-        strcasecmp(ext,"jsv")==0 ||
-        strcasecmp(ext,"avc")==0 ||
-        strcasecmp(ext,"26l")==0 ||
-        strcasecmp(ext,"jvt")==0 ) {
-            output = new EncodeOutputH264();
+    ext++;
+
+    if (codecName)
+        codec.assign(codecName);
+
+    if (!strcasecmp("AVC", codec.c_str()) || !strcasecmp(ext, "h264")
+        || !strcasecmp(ext, "264") || !strcasecmp(ext, "jsv")
+        || !strcasecmp(ext, "avc") || !strcasecmp(ext, "26l")
+        || !strcasecmp(ext, "jvt")) {
+        output = new EncodeOutputH264();
     }
-    else if((strcasecmp(ext,"ivf")==0) ||
-            (strcasecmp(ext,"vp8")==0)) {
-            output = new EncodeOutputVP8();
+    else if (!strcasecmp("VP8", codec.c_str())
+             || (!strcasecmp("VP8", codec.c_str()) && !strcasecmp(ext, "ivf"))
+             || !strcasecmp(ext, "vp8")) {
+        output = new EncodeOutputVP8();
     }
-    else if((strcasecmp(ext,"vp9")==0)) {
-            output = new EncodeOutputVP9();
+    else if (!strcasecmp("VP9", codec.c_str())
+             || (!strcasecmp("VP9", codec.c_str()) && !strcasecmp(ext, "ivf"))
+             || !strcasecmp(ext, "vp9")) {
+        output = new EncodeOutputVP9();
     }
-    else if((strcasecmp(ext,"jpg")==0) ||
-               (strcasecmp(ext,"jpeg")==0)) {
-               output = new EncodeStreamOutputJpeg();
-   }
-    else if((strcasecmp(ext,"h265")==0) ||
-               (strcasecmp(ext,"265")==0) ||
-               (strcasecmp(ext,"hevc")==0)) {
-               output = new EncodeOutputHEVC();
+    else if (!strcasecmp("JPEG", codec.c_str()) || !strcasecmp(ext, "jpg")
+             || !strcasecmp(ext, "jpeg")) {
+        output = new EncodeStreamOutputJpeg();
+    }
+    else if (!strcasecmp("HEVC", codec.c_str()) || !strcasecmp(ext, "h265")
+             || !strcasecmp(ext, "265") || !strcasecmp(ext, "hevc")) {
+        output = new EncodeOutputHEVC();
     }
     else
         return NULL;
 
-    if(!output->init(outputFileName, width, height)) {
+    if (!output->init(outputFileName, width, height)) {
         delete output;
         return NULL;
     }
--- a/tests/encodeinput.cpp
+++ b/tests/encodeinput.cpp
@@ -188,6 +188,9 @@ EncodeOutput* EncodeOutput::create(const char* outputFileName, int width , int h
             (strcasecmp(ext,"vp8")==0)) {
             output = new EncodeOutputVP8();
     }
+    else if((strcasecmp(ext,"vp9")==0)) {
+            output = new EncodeOutputVP9();
+    }
     else if((strcasecmp(ext,"jpg")==0) ||
                (strcasecmp(ext,"jpeg")==0)) {
                output = new EncodeStreamOutputJpeg();
@@ -237,6 +240,11 @@ const char* EncodeOutputVP8::getMimeType()
     return YAMI_MIME_VP8;
 }
 
+const char* EncodeOutputVP9::getMimeType()
+{
+    return YAMI_MIME_VP9;
+}
+
 const char* EncodeOutputHEVC::getMimeType()
 {
     return YAMI_MIME_HEVC;
@@ -254,30 +262,30 @@ void setUint16(uint8_t* header, uint16_t value)
     *h = value;
 }
 
-static void get_ivf_file_header(uint8_t *header, int width, int height,int count)
+void EncodeOutputVPX::getIVFFileHeader(uint8_t *header, int width, int height)
 {
     setUint32(header, YAMI_FOURCC('D','K','I','F'));
-    setUint16(header+4, 0);                     /* version */
-    setUint16(header+6,  32);                   /* headersize */
-    setUint32(header+8,  YAMI_FOURCC('V','P','8','0'));    /* headersize */
-    setUint16(header+12, width);                /* width */
-    setUint16(header+14, height);               /* height */
-    setUint32(header+16, 30);                    /* rate */
-    setUint32(header+20, 1);                    /* scale */
-    setUint32(header+24, count);                /* length */
-    setUint32(header+28, 0);                    /* unused */
+    setUint16(header+4, 0);                     // version
+    setUint16(header+6,  32);                   // headersize
+    setUint32(header+8,  m_fourcc);             // fourcc
+    setUint16(header+12, width);                // width
+    setUint16(header+14, height);               // height
+    setUint32(header+16, 30);                   // rate
+    setUint32(header+20, 1);                    // scale
+    setUint32(header+24, m_frameCount);         // framecount
+    setUint32(header+28, 0);                    // unused
 }
 
-bool EncodeOutputVP8::init(const char* outputFileName, int width , int height)
+bool EncodeOutputVPX::init(const char* outputFileName, int width , int height)
 {
     if (!EncodeOutput::init(outputFileName, width, height))
         return false;
     uint8_t header[32];
-    get_ivf_file_header(header, width, height, m_frameCount);
+    getIVFFileHeader(header, width, height);
     return EncodeOutput::write(header, sizeof(header));
 }
 
-bool EncodeOutputVP8::write(void* data, int size)
+bool EncodeOutputVPX::write(void* data, int size)
 {
     uint8_t header[12];
     memset(header, 0, sizeof(header));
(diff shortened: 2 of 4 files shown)


tests/encodeinput.h
```diff
@@ -147,16 +147,35 @@ class EncodeOutputH264 : public EncodeOutput
...
+class EncodeOutputVPX : public EncodeOutput {
 public:
-    EncodeOutputVP8();
-    ~EncodeOutputVP8();
-    virtual const char* getMimeType();
+    EncodeOutputVPX();
+    ~EncodeOutputVPX();
+    virtual const char* getMimeType() = 0;
     virtual bool write(void* data, int size);
+    void setFourcc(uint32_t fourcc) {m_fourcc = fourcc;};
```
xuguangxin: set and get should be protected

tests/encodeinput.h
```diff
@@ -147,16 +147,35 @@ class EncodeOutputH264 : public EncodeOutput
...
+    void getIVFFileHeader(uint8_t *header, int width, int height);
     int m_frameCount;
+    uint32_t m_fourcc;
+
+};
+
+class EncodeOutputVP8 : public EncodeOutputVPX {
+public:
+    EncodeOutputVP8();
+    ~EncodeOutputVP8(){};
```
xuguangxin: No need this. compiler will generate one for you

tests/encodeinput.cpp
```diff
@@ -176,29 +177,54 @@ EncodeOutput* EncodeOutput::create(const char* outputFileName, int width , int h
...
         return NULL;
     ext++;//h264;264;jsv;avc;26l;jvt;ivf
-    if(strcasecmp(ext,"h264")==0 ||
-        strcasecmp(ext,"264")==0 ||
-        strcasecmp(ext,"jsv")==0 ||
-        strcasecmp(ext,"avc")==0 ||
-        strcasecmp(ext,"26l")==0 ||
-        strcasecmp(ext,"jvt")==0 ) {
+
+    if (!strcasecmp(codecName, "AVC")) {
```
xuguangxin: We'd better make codecName as `optional `not `must`. If you do not set this we will guess from extension.If you set this we will ignore the extension. This code will break simple command line like this
yamencode -i xxx.264 -o xxx.265

tests/vppoutputencode.cpp
```diff
@@ -44,15 +44,17 @@ TranscodeParams::TranscodeParams()
...
+    if (!width || !height)
         if (!guessResolution(outputFileName, width, height))
             return false;
 
     m_fourcc = VA_FOURCC('N', 'V', '1', '2');
     m_width = width;
     m_height = height;
-    m_output.reset(EncodeOutput::create(outputFileName, m_width, m_height));
+    m_output.reset(EncodeOutput::create(outputFileName, m_width, m_height,
+                                        encParam.codec.c_str()));
```
xuguangxin: Use an uninitialized encParam?

tests/encodeinput.cpp
chivakker: This is kind of weird, if we're guessing based on the extension, then we should not give the option for codecName.  The problem then will be where two codecs can produce the same extension file, i.e. vpx both create ivf files.

tests/vppoutputencode.cpp
chivakker: The constructor will assign AVC.  Do you have command line examples where this code is exercised by the vpp?

tests/encodeinput.h
chivakker: Yes, thanks for pointing out, I'll fix.

tests/encodeinput.h
chivakker: yes, thanks for catching this.  I'll fix

chivakker:
I change the patch so that it can work on both scenarios when codecName is given and when it is not.  It should not change the behavior of the use cases mentioned for yamitranscode, yamiencode and yamivpp

tests/encodeinput.cpp
```diff
@@ -167,40 +167,47 @@ EncodeOutput::~EncodeOutput()
...
 {
-    EncodeOutput * output = NULL;
-    if(outputFileName==NULL)
+    EncodeOutput* output = NULL;
+    if (outputFileName == NULL)
         return NULL;
-    const char *ext = strrchr(outputFileName,'.');
-    if(ext==NULL)
+    const char* ext = strrchr(outputFileName, '.');
+    std::string codec(""); // needed for strcasecmp to work
```
xuguangxin: we can remove this if  you change 

> static EncodeOutput* create(const char* outputFileName, int width,  int height, const char* codecName = NULL);
to

> static EncodeOutput* create(const char* outputFileName, int width, int height, const char* codecName = NULL);

tests/encodeinput.cpp
```diff
@@ -167,40 +167,47 @@ EncodeOutput::~EncodeOutput()
...
-    if(strcasecmp(ext,"h264")==0 ||
-        strcasecmp(ext,"264")==0 ||
-        strcasecmp(ext,"jsv")==0 ||
-        strcasecmp(ext,"avc")==0 ||
-        strcasecmp(ext,"26l")==0 ||
-        strcasecmp(ext,"jvt")==0 ) {
-            output = new EncodeOutputH264();
+    ext++;
+
+    if (codecName)
```
xuguangxin: This no need too, if you change like above

tests/encodeinput.cpp
```diff
@@ -167,40 +167,47 @@ EncodeOutput::~EncodeOutput()
...
+        codec.assign(codecName);
+
+    if (!strcasecmp("AVC", codec.c_str()) || !strcasecmp(ext, "h264")
```
xuguangxin: we can change to !strcasecmp("AVC", codecName) if we set codecName default to "",
It no need extra variable and it no need copy the codecName to a new place

tests/encodeinput.cpp
chivakker: I'll check it but that will be c++11, won't it?

tests/encodeinput.cpp
```diff
@@ -167,40 +167,47 @@ EncodeOutput::~EncodeOutput()
...
+        || !strcasecmp(ext, "264") || !strcasecmp(ext, "jsv")
+        || !strcasecmp(ext, "avc") || !strcasecmp(ext, "26l")
+        || !strcasecmp(ext, "jvt")) {
+        output = new EncodeOutputH264();
+    }
+    else if ((!strcasecmp("VP8", codec.c_str()) && !strcasecmp(ext, "ivf"))
```
xuguangxin: This logical is not exactly as above. Previous logical is if user set "AVC", we will ignore extension. This logical is "VP8" is only for ivf.
We should ignore extension when user set the codecName, right?

tests/encodeinput.cpp
xuguangxin: Should not, It's just a point to a empty string "". Old c++ support this too.

xuguangxin:
1. Do we have plan to patch yamitranscode too?
2. Some code just a code sytle change, Please do not mix it with your main commit
you can refer to https://wiki.qt.io/Commit_Policy 8.1

> Make atomic commits. That means that each commit should contain exactly one self-contained change - do not mix unrelated changes, and do not create inconsistent states. Never "hide" unrelated fixes in bigger commits. Make coding style fixes only in exactly the lines which contain the functional changes, and comment fixes only when they relate to the change - the rest is for a separate commit.

tests/encodeinput.cpp
xuguangxin: Actually, this command reserved for .mkv or .mp4. Simpler solution for this is make .vp9 stand for vp9.  vp8 and ivf stand for vp8. :).   It is just our demo encoder; we can choose the extension and means. However, once you decide use the param, you had better make it consistency.

tests/vppoutputencode.cpp
xuguangxin: This is default value right? We do not need this since create function already have default value.

tests/encodeinput.cpp
chivakker: Ok, confirmed it is not c++11

tests/encodeinput.cpp
chivakker: yes, it will be changed.

tests/encodeinput.cpp
chivakker: yes, it will be changed.

tests/encodeinput.cpp
chivakker: It is the other way around, if ivf is the extension then codecName has to be set. And yes it is missing the case where extension is not on the command line but codecName is.  To be updated on next version.

chivakker:
>1. Do we have plan to patch yamitranscode too?
It is not in my plans, it should behave in the same way as before. 
>2. Some code just a code sytle change, Please do not mix it with your main commit you can refer to >https://wiki.qt.io/Commit_Policy 8.1

Ok

tests/encodeinput.cpp
chivakker: Actually this is the scenario:

Calling create without a codec name will result in codecName == ""
Calling create with codec but no codec given in command line results on codecName == NULL which requires a auxiliary std::string name to copy the "" for strcmpcase to work
Calling create with codec and codec given in command line results on codecName == "some name" which is ok.

Let me know where you want to handle the extra copy.  I think we can leave it here.  The other option is to add it to encodehelp.h when codec ==NULL

seanvk:
I will look for your updated patch set next week

tests/encodeinput.cpp
xuguangxin: To simply logical here you can make global variable codec default to "", then you have valid input for any case.

xuguangxin:
@chivakker  yamitranscode's code quality and performance is better than yamiencode. We had better remove yamiencode after all things supported in yamitranscode.  Maybe we can discuss yamitranscode support plan in weekly.

tests/encodeinput.cpp
```diff
@@ -167,40 +167,49 @@ EncodeOutput::~EncodeOutput()
...
+        codec.assign(codecName);
+
+    if (!strcasecmp("AVC", codec.c_str()) || !strcasecmp(ext, "h264")
+        || !strcasecmp(ext, "264") || !strcasecmp(ext, "jsv")
+        || !strcasecmp(ext, "avc") || !strcasecmp(ext, "26l")
+        || !strcasecmp(ext, "jvt")) {
+        output = new EncodeOutputH264();
+    }
+    else if (!strcasecmp("VP8", codec.c_str())
+             || (!strcasecmp("VP8", codec.c_str()) && !strcasecmp(ext, "ivf"))
```
lizhong1008: why we need this line?  !strcasecmp("VP8", codec.c_str() is enough.

tests/encodeinput.cpp
```diff
@@ -167,40 +167,49 @@ EncodeOutput::~EncodeOutput()
...
+             || !strcasecmp(ext, "vp8")) {
+        output = new EncodeOutputVP8();
+    }
+    else if (!strcasecmp("VP9", codec.c_str())
+             || (!strcasecmp("VP9", codec.c_str()) && !strcasecmp(ext, "ivf"))
```
lizhong1008: redundant line

tests/encodeinput.cpp
chivakker: if command line doesn't have codeName "-c 'VPx'" (with x == 8.9) and ivf is the extension then there's no way to know which codec to launch

tests/encodeinput.cpp
chivakker: same explanation as above.

tests/encodeinput.cpp
xuguangxin: This will force user assign a codec for ivf. This wired since other codec do not have this requirements. Do you think it is ok to make vp8 as default codec for ivf, and if user assign a “vp9” to codec, we will encode vp9 for them?
The logical should like this
```
else if (!strcasecmp("VP8", codec.c_str())
            || (!strcasecmp("", codec.c_str()) && !strcasecmp(ext, "ivf")) {
	xxx
} else if (!strcasecmp("VP9", codec.c_str())
	|| !strcasecmp(ext, "ivf")) {
	xxxx.
}
```

tests/encodeinput.cpp
chivakker: Can you consider the current solution as a valid alternative to declaring a global variable?

tests/encodeinput.cpp
chivakker: what are the use cases that require ivf to be defaulted to vp8?  If someone requires vp8 or vp9 with only the extension then use .vp8 and .vp9.

lizhong1008: @chivakker yeah, I know what you mean, but if you set -C vp8, line 191 is enough, right? Anyway, as you logic, line 192 is unnecessary.

lizhong1008: We don't need to guess to much, vp9 may be more and more popular than vp8. If user use ivf, he should assign a codec type.

chivakker: The choice of extension shouldn't be based on popularity.  In effect ivf is the only extension with this problem so far but that's what our application does, it puts vp8 and vp9 content with ivf headers in a file. I think this is a minor problem and if there's a real use case that needs vp8 and ivf assumption I would like to know it

xuguangxin: Agree with Daniel, we should not choose behaviour base on popularity, Popularity will change time to time. 
Hi @chivakker , this code will make  simple command “yamidecode –i xxx.264 –o ttt.ivf” failed, and we cannot provide any descriptive information here.  Who knows we need provide other –c command  to make it work before he dive into the code?

chivakker: Hey @xuguangxin, I am assuming you meant to say "yamitranscode -i xxx.264 -o ttt.ivf".  In this case and by adding vp9 encoder support we need to de-generalize the ivf extension.  My suggestion would be to change all those use cases with the extension associated with the codec.  IIUC, today all .ivf examples mean vp8 so the scripts can be renamed to do "yamitranscode -i xxx.264 -o ttt.vp8" and include new use cases for vp9 like "yamitranscode -i xxx.264 -o ttt.vp9"

Without the first check on the vp8/vp8 if statements that would not be possible.

I understand yamitranscode should be able to operate without much information on the command line.

xuguangxin: yamiencode can accept .264 as input too. However, it is much slow than yamitranscode. You can try it ;).  How about make ivf default for vp8, if user assign "-c vp9 xx2.ivf" we output vp9 to ivf. Use this approach, it have two benefits:
1.	QA do not need change script.
2.	It works when we assign or not assign –C param in any case.

chivakker: > 2. It works when we assign or not assign –C param in any case.

this patch will work when -C is not given when using vp8 and vp9 extensions.  For that I need to understand the impact on QA.  I'll ping the people and comment back.

lizhong1008: @chivakker , I agree with you, we don't need to suppose vp8 codec type for ivf, it will force a vp8 encoding but maybe user want a vp9 encoder. I recommend we use xxx_vp9.ivf or xxx_vp8.ivf, it should be better than xxx.vp8/xxx.vp9, becasue we can know it is ivf format not a pure vp8/vp9 bitstream.

lizhong1008: @chivakker another question, any concern about "line 192 is unnecessary."?

xuguangxin: Hi @lizhong1008 :
This will introduce unnecessary complexity to our encoder codec guess logical. You can look at EncodeOutput::create. We only guess from extension yet. Actually for Daniel’s patch, you can add `-c vp9` to get vp9 encoded .ivf file. Daniel and I do not think ivf should for vp8 only.
The main difference between Daniel and me is I think we should treat vp8 as default codec format for ivf. i.e. 
1.	If you use `yamiencode –i xxx.264 –o ttt.ivf`, we will encode vp8 for you. 
2.	If you use `yamiencode –i xxx.264 –c vp9 –o ttt.ivf`, we will encode vp9 for you.
Daniel suggest we failed and report an error to user in case 1. 
What do you prefer?

lizhong1008: 1. I agree with Daniel we should report an error to user for case 1.  User should set codec type when this error happens. It is unnecessary to set vp8 as default, because it will waste time to encoding a vp8 bitstream if user actually want a vp9 encoder. And it brings another question, why we set vp8 as default, not vp9? Yeah, case 1 failed need QA change their scripts, let's get feedback from them, I think this effort with litter burden and can be ignored.
2. Yes, EncodeOutput::create only guess extension right now, but could we add strstr() to check the whole string "outputFileName"? seems no much complexity.

xuguangxin: 1. Ok, I can accept if you all agree on this; it should be easy for QA to change script.
2.  The output file name is provide by user, _vp8.ivf looks like magic to them. Moreover "-c vp8" can assign encode codec for ivf. It’s more straightforward than _vp8.ivf

lizhong1008: I don't mean we should force user to use such a name like XXX_vp8.ivf, but if he already used such a name, we should guess he want a vp8 encoder, even without "-c vp8" assigned, right?

tests/encodeinput.cpp
xuguangxin: We already have a global variable here https://github.com/01org/libyami-utils/blob/master/tests/encodehelp.h#L28, But it default to NULL. I mean, if you set this default to “”
It will make code more cleaner.Just a suggestion ;)

chivakker:
I want to start fixing the problems with yamitranscode (with vp9) and it would help if we can merge this.  Let me know what is holding the merge of this patch. I am not willing to advertise yamitranscode as an option for vp9 encoder 'till those problems are addressed.

tests/encodehelp.h
```diff
@@ -59,8 +59,8 @@ static void print_help(const char* app)
     printf("   -o <coded file> optional\n");
     printf("   -b <bitrate: kbps> optional\n");
     printf("   -f <frame rate> optional\n");
-    printf("   -c <codec: AVC|VP8|JPEG> Note: not support now\n");
-    printf("   -s <fourcc: NV12|IYUV|YV12> Note: not support now\n");
+    printf("   -c <codec: AVC|VP8|VP9|JPEG>\n");
```
lizhong1008: We need also add HEVC support.

xuguangxin:
Hi @wudping 
Please help to verify this. to make sure we can encode a vp9 stream.
thanks.

wudping:
okay, @xuguangxin , you are welcome.

xuguangxin:
+1

wudping:
verify + 1

lizhong1008:
merged as 018276c62d62950175b70a68d2e78c40334db1b9

PR #17, opened 2016-08-18
v4l2decode: support wayland display

YuJiankang:
add wayland support to v4l2 decoder

### DIFF
--- a/tests/v4l2decode.cpp
+++ b/tests/v4l2decode.cpp
@@ -36,6 +36,8 @@
 #if ANDROID
 #include <gui/SurfaceComposerClient.h>
 #include <va/va_android.h>
+#elif __ENABLE_WAYLAND__
+#include <wayland-client.h>
 #else
 #include "./egl/gles2_help.h"
 #endif
@@ -200,15 +202,12 @@ static std::vector<uint8_t*> inputFrames;
 static std::vector<struct RawFrameData> rawOutputFrames;
 
 static FILE* outfp = NULL;
-#ifndef ANDROID
+
+#if (!defined(ANDROID) && !defined(__ENABLE_WAYLAND__))
 static Display * x11Display = NULL;
 static Window x11Window = 0;
-#endif
-#ifndef ANDROID
 static EGLContextType *eglContext = NULL;
 static std::vector<EGLImageKHR> eglImages;
-#endif
-#ifndef ANDROID
 static std::vector<GLuint> textureIds;
 #endif
 static bool isReadEOS=false;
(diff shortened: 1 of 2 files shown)


tests/v4l2decode.cpp
```diff
@@ -330,6 +329,153 @@ static bool displayOneVideoFrameAndroid(int32_t fd, int32_t index)
 
     return true;
 }
+
+#elif __ENABLE_WAYLAND__
```
halleyzhao: 不同平台的接口类似，用类来实现可能更清晰一下。

tests/v4l2decode.cpp
```diff
@@ -434,7 +581,7 @@ int main(int argc, char** argv)
     int32_t ioctlRet = -1;
     YamiMediaCodec::CalcFps calcFps;
 
-#if __ENABLE_X11__
+#if (defined (__ENABLE_X11__) && !defined(__ENABLE_WAYLAND__))
```
halleyzhao: 将不同的平台作为互斥的选项可能会方便一些。 --enable-platform=x11/wayland/android

tests/v4l2decode.cpp
```diff
@@ -330,6 +329,153 @@ static bool displayOneVideoFrameAndroid(int32_t fd, int32_t index)
...
+    output((struct wl_buffer *)(buf->m.userptr));
+    memset(&buffer, 0, sizeof(buffer));
+    buffer.type = V4L2_BUF_TYPE_VIDEO_CAPTURE_MPLANE;
+    buffer.index = (buf->index) ? (buf->index - 1): (outputQueueCapacity - 1);
+    ioctlRet = SIMULATE_V4L2_OP(Ioctl)(fd, VIDIOC_QBUF, &buffer);
+    ASSERT(ioctlRet != -1);
+    return true;
+}
 #else
 bool displayOneVideoFrameEGL(int32_t fd, int32_t index)
```
halleyzhao: 这个函数名称可以改一下了。

tests/v4l2decode.cpp
xuguangxin: Agree, we need refact the code, not it's bit of mess here.

tests/v4l2decode.cpp
```diff
@@ -36,6 +36,8 @@
 #if ANDROID
 #include <gui/SurfaceComposerClient.h>
 #include <va/va_android.h>
+#elif __ENABLE_WAYLAND__
+#include <va/va_wayland.h>
```
halleyzhao: do we really use va_wayland.h here?

tests/v4l2decode.cpp
```diff
@@ -537,7 +684,17 @@ int main(int argc, char** argv)
     }
     ioctlRet = SIMULATE_V4L2_OP(Ioctl)(fd, VIDIOC_S_FMT, &format);
     ASSERT(ioctlRet != -1);
-
+#if __ENABLE_WAYLAND__
```
halleyzhao: how about move it above to be parallel with XOpenDisplay().

tests/v4l2decode.cpp
```diff
@@ -715,7 +871,17 @@ int main(int argc, char** argv)
...
     }
 #endif
-
+#if __ENABLE_WAYLAND__
+    struct v4l2_buffer buffer;
+    //queue buffs
+    for (uint32_t i = 0; i < outputQueueCapacity; i++) {
+        memset(&buffer, 0, sizeof(buffer));
+        buffer.type = V4L2_BUF_TYPE_VIDEO_CAPTURE_MPLANE;
+        buffer.index = i;
```
halleyzhao: usually buffer.memory is required for VIDIOC_QBUF/VIDIOC_DQBUF.

tests/v4l2decode.cpp
```diff
@@ -715,7 +871,17 @@ int main(int argc, char** argv)
         }
     }
 #endif
-
+#if __ENABLE_WAYLAND__
```
halleyzhao: in fact, it does same as following "#ifndef ANDROID   takeOneOutputFrame(fd, i)"

tests/v4l2decode.cpp
```diff
@@ -330,6 +329,153 @@ static bool displayOneVideoFrameAndroid(int32_t fd, int32_t index)
...
+    return vaPutSurfaceWayland(buffer, &srcRect, &dstRect);
+}
+
+static bool displayOneVideoFrameWayland(int32_t fd, struct v4l2_buffer  *buf)
+{
+    int32_t ioctlRet = -1;
+    struct v4l2_buffer buffer;
+
+    output((struct wl_buffer *)(buf->m.userptr));
+    memset(&buffer, 0, sizeof(buffer));
```
halleyzhao: in fact, takeOneOutputFrame() does VIDIOC_QBUF.
we needn't QBUF here, nor for displayOneVideoFrameAndroid. (just be careful that buffer index may change).

tests/v4l2decode.cpp
halleyzhao: ok, buffer.memory is useless for wl_buffer type. we can ignore it.

halleyzhao: how about something like V4L2_MEMORY_DRIVERPTR? then we needn't make exception for the memory type elsewhere.

tests/v4l2decode.cpp
```diff
@@ -330,6 +329,153 @@ static bool displayOneVideoFrameAndroid(int32_t fd, int32_t index)
...
+    buffer.type = V4L2_BUF_TYPE_VIDEO_CAPTURE_MPLANE;
+    buffer.index = (buf->index) ? (buf->index - 1): (outputQueueCapacity - 1);
```
halleyzhao: why minus one?

tests/v4l2decode.cpp
```diff
@@ -330,6 +329,153 @@ static bool displayOneVideoFrameAndroid(int32_t fd, int32_t index)
...
+    SharedPtr<wl_compositor>     compositor;
+    SharedPtr<wl_shell>          shell;
+    SharedPtr<wl_shell_surface>  shell_surface;
+    SharedPtr<wl_surface>        surface;
+};
+
+static struct display waylandDisplay;
+static bool redrawPending = false;
+
+struct VADisplayTerminator {
```
halleyzhao: is it really used? VADisplayTerminator

tests/v4l2decode.cpp
```diff
@@ -200,15 +202,12 @@ static std::vector<uint8_t*> inputFrames;
 static std::vector<struct RawFrameData> rawOutputFrames;
 
 static FILE* outfp = NULL;
-#ifndef ANDROID
+
+#if (!defined(ANDROID) && !defined(__ENABLE_WAYLAND__))
```
xuguangxin: should be #if !__ENABLE_X11__

tests/v4l2decode.cpp
```diff
@@ -330,6 +329,153 @@ static bool displayOneVideoFrameAndroid(int32_t fd, int32_t index)
...
+        if (!d->shell_surface)
+            return false;
+        wl_shell_surface_set_toplevel(d->shell_surface.get());
+    }
+    return true;
+}
+
+bool createWaylandDisplay()
+{
+    SharedPtr<VADisplay> display;
```
halleyzhao: SharedPtr<VADisplay> display is not used.

tests/v4l2decode.cpp
YuJiankang: yes, how about merge it firstly? and we take some time to refine the code in the future.

tests/v4l2decode.cpp
YuJiankang: minus one, means the we can reuse the last surface for output.

tests/v4l2decode.cpp
YuJiankang: ok, I will try move it to be parallel with XOpenDisplay().

tests/v4l2decode.cpp
YuJiankang: yes, wayland related structure and interface are included by this header file.

tests/v4l2decode.cpp
YuJiankang: yes, I will remove it, thanks

tests/v4l2decode.cpp
YuJiankang: yes, I will remove it, thanks

tests/v4l2decode.cpp
halleyzhao: if wayland only, then wayland, not va/wayland.

tests/v4l2decode.cpp
```diff
@@ -330,6 +329,153 @@ static bool displayOneVideoFrameAndroid(int32_t fd, int32_t index)
...
+    if (!ensureWindow(dstRect->width, dstRect->height))
+        return false;
+
+    wl_surface_attach(d->surface.get(), buffer, 0, 0);
+    wl_surface_damage(d->surface.get(), dstRect->x,
+		dstRect->y, dstRect->width, dstRect->height);
+    wl_display_flush(d->display.get());
+    redrawPending = true;
+    callback = wl_surface_frame(d->surface.get());
+    wl_callback_add_listener(callback, &frame_callback_listener,&redrawPending);
```
halleyzhao: frame_callback_listener is local variable, i'm not sure whether it is ok to set it to a callback listener.

tests/v4l2decode.cpp
```diff
@@ -330,6 +329,153 @@ static bool displayOneVideoFrameAndroid(int32_t fd, int32_t index)
...
+    wl_surface_commit(d->surface.get());
+    return true;
```
halleyzhao: seems, wl_buffer should be destroyed on a callback. otherwise, some mem leak.

tests/v4l2decode.cpp
YuJiankang: ok, I will try include wayland-client.h, thanks

tests/v4l2decode.cpp
YuJiankang: yes, may be a problem, I will try to destroy the wl_buffer

tests/v4l2decode.cpp
```diff
@@ -330,6 +329,153 @@ static bool displayOneVideoFrameAndroid(int32_t fd, int32_t index)
 
     return true;
 }
+
+#elif __ENABLE_WAYLAND__
+struct display {
+    SharedPtr<wl_display>        display;
```
halleyzhao: not so good, you can't decide the order to destroy these object.
if wl_display is destroyed before wl_surface, there will be exception.

halleyzhao:
it's ok to merge first, then does some refine work later.

xuguangxin:
+1, thanks Halley for review. 
merged as 896be8bd26c119d36da0fe58cafb280a1499806b

PR #18, opened 2016-08-19
Fix jpeg conformance.

xuguangxin:
Previous code will convert all output to i420, using software, it assume all decoder output is NV12. However, this not true for jpeg, jpeg will output many formats like IMC3, 422H, 422V, 444P etc. This patch and related patch in libyami will skip the conversion if user do not assign one. So we can compare the yuv with software decoder.

### DIFF
--- a/testscripts/ssim_test.py
+++ b/testscripts/ssim_test.py
@@ -1,0 +1,118 @@
+#!/usr/bin/env python3
+
+#This script will compare ssim value between libyami and ffmpeg.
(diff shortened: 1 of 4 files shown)


uartie:
lgtm :+1:

testscripts/ssim_test.py
```diff
@@ -0,0 +1,111 @@
+#!/usr/bin/env python3
```
chivakker: would you mind adding a description of this script as a comment here?  It can also include examples and maybe a help page?

xuguangxin: Good suggestion, added bellow

lizhong1008:
merged and closed.

PR #19, opened 2016-08-24
Makefile: compile decode|encode|vpp|transcode if --enablecapi

wudping:
Compile yami[decode|encode|vpp|transcode] together with --enablecapi

To fix the bug: 
Enabling capi and tests should not skip yami[decode|encode|vpp|transcode] compilation 


xuguangxin:
-1, our c api and c++ api have different memory layout, final address this problem need remove the constructor from decoder/encoder.

wudping:
@xuguangxin , okay, I will update.

xuguangxin:
Hi @wudping 
Any update on this?

wudping:
Hi @xuguangxin ,
I am working on this now.  The yamidecode can use --enable-capi option, but yamitranscode is still need more effort. I will push the patch in this week.

xuguangxin:
Hi @wudping 
We are not suppose have --enable-capi in compile time any more. We should have --capi option in yamidecode and yamitranscode in run time.

wudping:
okay @xuguangxin , I get it.

tests/decode.cpp
```diff
@@ -33,25 +30,35 @@
...
+            fprintf(stderr, "createInput failed, cannot convert VppInput to VppInputDecode\n");
+            return input;
+        }
+        if (!inputDecode->config(*display)) {
+            input.reset();
+            fprintf(stderr, "VppInputDecode config failed.\n");
+            return input;
+        }
+    }else{
+        SharedPtr<VppInputDecode> inputDecode = std::tr1::dynamic_pointer_cast<VppInputDecode>(input);
```
xuguangxin: Totally duplicate code with if branch, Do not repeat yourself is important  please refer to https://en.wikipedia.org/wiki/Don%27t_repeat_yourself

tests/decodehelp.cpp
```diff
@@ -48,20 +49,28 @@ static void printHelp(const char* app)
     printf("      3: texture: export video frame as drm name (RGBX) + texture from drm name\n");
     printf("      4: texture: export video frame as dma_buf(RGBX) + texutre from dma_buf\n");
     printf("      5: texture: export video frame as dma_buf(NV12) + texture from dma_buf. not implement yet\n");
+    printf("  --use-capi: use the codec capi to encode or decode, default(false)\n");
```
xuguangxin: just name it as --capi, and group it after -m.

tests/vppoutputencode.cpp
```diff
@@ -126,6 +126,11 @@ static void setEncodeParam(const SharedPtr<IVideoEncoder>& encoder,
     encoder->setParameters(VideoConfigTypeAVCStreamFormat, &streamFormat);
 }
 
+VppOutputEncode::VppOutputEncode()
+{
+    m_outputBuffer.format = OUTPUT_BUFFER_LAST;
```
xuguangxin: why we need change this?

tests/decodehelp.cpp
wudping: okay, thanks @xuguangxin .

tests/decode.cpp
wudping: thanks @xuguangxin , but the member function config() is not a virtual function.

tests/vppoutputencode.cpp
wudping: m_outputBuffer.format has a initial value before this patch when --disable-capi
`#ifndef __ENABLE_CAPI__		
      VideoEncOutputBuffer():data(0), bufferSize(0), dataSize(0)		
     , remainingSize(0), flag(0),  **format(OUTPUT_BUFFER_LAST)**, timeStamp(0) {		
     };		
 #endif`

It's okay to delete this piece of code. I had tested this patch without this.

tests/decode.cpp
xuguangxin: We can make code better like this:
```
if(para.useCAPI) {
    SharedPtr<VppInputDecodeCapi> inputDecode = std::tr1::dynamic_pointer_cast<VppInputDecodeCapi>(input);		       
    if (inputDecode && inputDecode->config(*display))
          return input;
} else {
....
}
input.reset();
fprintf(stderr, "invalid input file.\n");
```

tests/decodehelp.cpp
```diff
@@ -92,6 +101,11 @@ bool processCmdLine(int argc, char** argv, DecodeParameter* parameters)
         case 'o':
             outputFile = optarg;
             break;
+        case 0:
+            switch (option_index) {
+            case 1:
+                parameters->useCAPI = strncasecmp(optarg, "true", 4) == 0;
+                break;
```
xuguangxin: wired, is it build able

tests/decode.cpp
wudping: okay, thanks @xuguangxin .

tests/decodehelp.cpp
wudping: yes, it is. I have included the head file "#include <getopt.h>".

xuguangxin: ok, i see you removed the { at line 73, could you not remove it?

tests/decodehelp.cpp
```diff
@@ -92,6 +101,11 @@ bool processCmdLine(int argc, char** argv, DecodeParameter* parameters)
         case 'o':
             outputFile = optarg;
             break;
+        case 0:
+            switch (option_index) {
```
xuguangxin: this { will match with L112, and they are not in same indent level.

wudping: oh, you are right, I am sorry. Thanks @xuguangxin, I will update it.

egl/egl_util.h
```diff
@@ -19,7 +19,7 @@
 #include <EGL/egl.h>
 #include <EGL/eglext.h>
 
-#include "VideoCommonDefs.h"
+#include <VideoCommonDefs.h>
```
xuguangxin: Please use Libyami.h and LibyamiC.h for all c++ and c place.

wudping:
covered by regression test

PR #28, opened 2016-09-08
Compiler fixes + TR1 isolation

uartie:
fixes for gcc 6.x and TR1 issues that will become more apparent with libaymi change https://github.com/01org/libyami/pull/573

### DIFF
--- a/tests/vppoutputencode.cpp
+++ b/tests/vppoutputencode.cpp
@@ -63,7 +63,7 @@ bool VppOutputEncode::init(const char* outputFileName, uint32_t /*fourcc*/, int 
     m_width = width;
     m_height = height;
     m_output.reset(EncodeOutput::create(outputFileName, m_width, m_height));
-    return m_output;
+    return bool(m_output);
 }
 
 void VppOutputEncode::initOuputBuffer()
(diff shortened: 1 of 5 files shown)


tests/vppoutputencode.cpp
```diff
@@ -63,7 +63,7 @@ bool VppOutputEncode::init(const char* outputFileName, uint32_t /*fourcc*/, int
     m_width = width;
     m_height = height;
     m_output.reset(EncodeOutput::create(outputFileName, m_width, m_height));
-    return m_output;
```
xuguangxin: this not needed you can refer to http://en.cppreference.com/w/cpp/memory/shared_ptr/operator_bool

tests/yamitranscode.cpp
```diff
@@ -300,7 +300,7 @@ class TranscodeTest
             return false;
         }
         m_allocator = createAllocator(m_output, m_display, m_cmdParam.m_encParams.ipPeriod);
-        return m_allocator;
```
xuguangxin: it's not needed too

tests/vppoutputencode.cpp
xuguangxin: My mistake, Now i know c++ 11 do not allow this

tests/yamitranscode.cpp
xuguangxin: my mistake.

tests/vppoutputencode.cpp
uartie: Right.  C++11 share_ptr bool operator is defined as `explicit` which is why we need this.

PR #29, opened 2016-09-13
do not use ::std::tr1 explicitly

uartie:
Commit 8dc7783d1da6c reversed some of the changes in
commit 4ae3688ca7eb to use DynamicPointerCast instead
of std::tr1::dynamic_pointer_cast.  Libyami encapsulates
the use of ::std::tr1:: vs. ::std:: based on compiler.
Thus, use libyami's interface for this in decode.cpp.

Also, use libyami's SharedPointerCast in grid.cpp.


xuguangxin:
+1

xuguangxin:
Thanks for patch.

PR #34, opened 2016-09-14
yamiencode: mark a force keyframe according to the intraPeriod

### LINKED ISSUE #33: keyFrame management with vpx encoders
Application (yamiencode) can call encode and force the request of a key frame.  This way the component doesn't need to keep track of intraPeriod and the application can do it upon request.

chivakker:
application can request the libyami component a keyFrame on demand.
The application is requesting it based on the intraPeriod value and
libyami component will act upon.

Do this only for VPX encoding and let h.26x decide if they want to
incorporate this change later

Signed-off-by: Daniel Charles <daniel.charles@intel.com>

Fixes #33 
Requires: https://github.com/01org/libyami/pull/588

### DIFF
--- a/tests/encode.cpp
+++ b/tests/encode.cpp
@@ -34,6 +34,11 @@
 
 using namespace YamiMediaCodec;
 
+inline bool isVPX(const char* codec)
+{
+    return !strcmp(codec, YAMI_MIME_VP9) || !strcmp(codec, YAMI_MIME_VP8);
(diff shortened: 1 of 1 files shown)


chivakker:
Abandoning this implementation as there are other aspects to consider and this is not the correct way of forcing a keyFrame.  Thanks @xuguangxin for pointing that out.

PR #35, opened 2016-09-18
transcode: add input/output width/height setting

lizhong1008:
1. It to fix https://github.com/01org/libyami/issues/578
2. Output resolution setting is optional, when it is not set, it will be
same as input resolution. And if it is different from input resolution,
scaling happens.

Signed-off-by: Zhong Li <zhong.li@intel.com>


xuguangxin:
merged as f3c21ae5ed2ee80befe54b41b6907a6de147d8f1

PR #36, opened 2016-09-19
jpg_ssim: add suffix '.mjpg' into ssim_test.py

wudping:
add jpeg suffix '.mjpg' into ssim_test.py,
so that ssim_test.py can test the conformance
of those jpeg clips.

### DIFF
--- a/testscripts/ssim_test.py
+++ b/testscripts/ssim_test.py
@@ -23,6 +23,8 @@ def isCandidate(f):
         , ".mpeg"
         , ".mpg"
         , ".mpeg2"
+        , ".mjpg"
+        , ".mjpeg"
     ]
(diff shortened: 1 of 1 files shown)


testscripts/ssim_test.py
```diff
@@ -23,6 +23,7 @@ def isCandidate(f):
         , ".mpeg"
         , ".mpg"
         , ".mpeg2"
+        , ".mjpg"
```
xuguangxin: please add .mjpeg too

wudping: okay, thanks for your suggestion @xuguangxin .

wudping:
updated according to @xuguangxin's advises.

PR #38, opened 2016-09-22
v4l2: return status value in main function

xuguangxin:


xuguangxin:
@lizhong1008 , help to review

tests/v4l2decode.cpp
```diff
@@ -1031,5 +1031,6 @@ int main(int argc, char** argv)
 #endif
 
     fprintf(stdout, "decode done\n");
+    return 0;
```
lizhong1008: do we need change int main to void main?

xuguangxin: We can't, this will return -1, status code is a communication between application and shell

```
    input = DecodeInput::create(params.inputFile);
    if (input==NULL) {
        ERROR("fail to init input stream\n");
        return -1;
    }
```

lizhong1008: ok， let‘s keep to return 0

lizhong1008:
+1

chivakker:
lgtm :+1:

PR #39, opened 2016-09-22
yamitransocde: implement flush for 264/265 to fix #598

wudping:
In the case that when we pass value 5 via --ipperiod to yamiencode, and
yamiencode conserved 2 YUV picture buffers(I B B) as B frame, the file
ends up; yamiencode does not encode until it gets a P/I frame; so those
two B frames are lost.
Added flush() function into class VppOutput, to make codec(264/265) en-
code the left B frames.

### DIFF
--- a/tests/vppoutputencode.cpp
+++ b/tests/vppoutputencode.cpp
@@ -176,6 +176,8 @@ bool VppOutputEncode::output(const SharedPtr<VideoFrame>& frame)
             fprintf(stderr, "encode failed status = %d\n", status);
             return false;
         }
+    }else {
+        m_encoder->flush();
     }
     do {
         status = m_encoder->getOutput(&m_outputBuffer, drain);
--- a/tests/yamitranscode.cpp
+++ b/tests/yamitranscode.cpp
@@ -342,7 +342,11 @@ public:
             if(count >= m_cmdParam.frameCount)
                 break;
         }
+        src.reset();
+        m_output->output(src);
+
         fps.log();
+
         return true;
     }
 private:


tests/vppoutputencode.cpp
```diff
@@ -186,3 +186,19 @@ bool VppOutputEncode::output(const SharedPtr<VideoFrame>& frame)
     return true;
 
 }
+
+bool VppOutputEncode::flush()
```
xuguangxin: could we move this  VppOutputEncode::~VppOutputEncode::()?
you may need check m_encoder and m_output  before use it.

tests/vppoutputencode.cpp
```diff
@@ -186,3 +186,19 @@ bool VppOutputEncode::output(const SharedPtr<VideoFrame>& frame)
...
+{
+    Encode_Status status = ENCODE_SUCCESS;
+    bool drain = false;
+
+    m_encoder->flush();
+    do {
+        status = m_encoder->getOutput(&m_outputBuffer, drain);
```
xuguangxin: do not duplicate this code, could we use output(NULL) for flush?

tests/vppoutputencode.cpp
wudping: okay, thanks, @xuguangxin .

tests/vppoutputencode.cpp
wudping: okay, thanks, @xuguangxin

wudping:
updated according to comments, please review.

chivakker:
lgtm :+1: but please make sure that libyami fixes the base encoder class.

xuguangxin:
merged as 34f4dff01fa2fdd58c93e2220794d737bca926db

PR #42, opened 2016-09-23
tests: encode, flush frame at eos

xuguangxin:
When we have b frame, encoder will cache some frame to adjust reference order.
We need flush them, else we will lost frame


lizhong1008:
+1

xuguangxin:
@wudping , Please help to verify.

wudping:
sorry, @xuguangxin , I am in the meeting.

wudping:
verify + 1

PR #45, opened 2016-10-01
[v4l2]:use return value EAGAIN instread of errno

halleyzhao:


tests/v4l2decode.cpp
```diff
@@ -590,15 +590,21 @@ int main(int argc, char** argv)
...
-            if (errno != EINVAL) {
-                // EINVAL means we haven't seen sufficient stream to decode the format.
-                INFO("ioctl() failed: VIDIOC_G_FMT, haven't get video resolution during start yet, waiting");
-            }
-        } else {
+        ioctlRet = SIMULATE_V4L2_OP(Ioctl)(fd, VIDIOC_G_FMT, &format) ;
+        if (ioctlRet == 0)
             break;
+
+        #ifdef YUNOS
```
chivakker: Please remove this pre-processor check and the whole #else case.  This is the same comment as with previous patches.

tests/v4l2decode.cpp
```diff
@@ -590,15 +590,21 @@ int main(int argc, char** argv)
...
+        #else
+        if (errno == EINVAL) {
+            // EINVAL means we haven't seen sufficient stream to decode the format.
+        #endif
+            INFO("ioctl() failed: VIDIOC_G_FMT, haven't get video resolution during start yet, waiting");
+        } else {
+            WARNING("ioctl fail to get VIDIOC_G_FMT");
         }
-        usleep(50);
+        usleep(5000);
```
chivakker: Can a good explanation for increment the sleep time here be given?

xuguangxin:
Hi @halleyzhao 
Any update on this?

tests/v4l2decode.cpp
halleyzhao: updated.

tests/v4l2decode.cpp
chivakker: Can you add the explanation as a comment in the code? sleeps are not recommended mostly.

halleyzhao:
add comment “// decoding runs in another thread, wait until video info is got before continue”

( though I think the code itself is straightforward enough. when there is EAGAIN, you have to sleep/wait for single thread client).

MicheleL-Intel:
What is the current status?  Several changes were proposed:
1) Increment sleep count (agree with Daniel, sleep in general is not recommended).  Presume this change did not take place since there is no commit.  Please confirm.
2) Changed macro definition (submitted)
3) use return value EAGAIN instread of errno (submitted)

If you feel (2) & (3) are the only changes needed, please rebase and resubmit so we can close.  Thank you.

tests/v4l2decode.cpp
MicheleL-Intel: Suggest using other means such as a callback to indicate when resolution info is available, rather than hardcoding the wait time.  This is not efficient especially because 'usleep' is called within an infinite 'while' loop.  So every time you run thru the loop, it blocks for 5 ms.

xuguangxin: Hi Michele:
This just a sample application for v4l2decoder. To simplify the thread logical, it waits here until he gets the format change information. This not harms too much to performance, it just sleep 5 ms at the start. The real caller will call this function after he get notification from the interrupt thread like https://chromium.googlesource.com/chromium/src/media/+/master/gpu/v4l2_video_decode_accelerator.cc#2054
.

uartie:

PR #46, opened 2016-10-08
makefile: add top_srcdir into CPPFLAGS for yocto

wudping:
to integrate libyami into yocto, we should add top_srcdir into cppflags
to make it compiled well.


examples/Makefile.am
```diff
@@ -17,6 +17,7 @@ AM_CFLAGS = \
 	-I../tests \
```
chivakker: Can you do something like this:

extra_Includes = \
-I $(top_srcdir) \ #to compile within yocto

and then you add the extra_includes wherever you need.  

That way can help us identify in the future why this was added and maybe modify it.

wudping: okay, thanks @chivakker

examples/Makefile.am
```diff
@@ -73,21 +74,24 @@ VPP_INPUT_LIBS = \
...
 VPP_INPUT_CFLAGS = \
 	$(LIBVA_CFLAGS) \
+	-I$(top_srcdir) \
 	$(NULL)
 
 if ENABLE_X11
 simpleplayer_LDADD = $(YAMI_DECODE_LIBS)
 simpleplayer_LDFLAGS = $(LIBYAMI_CFLAGS)
 simpleplayer_SOURCES = simpleplayer.cpp $(DECODE_INPUT_SOURCES)
+simpleplayer_CPPFLAGS = $(LIBYAMI_CFLAGS) $(LIBDRM_CFLAGS) $(AM_CPPFLAGS) -I$(top_srcdir)
```
xuguangxin: Why we need this?We already added them to AM_CFLAGS/AM_CPPFLAGS

examples/Makefile.am
```diff
@@ -73,21 +74,24 @@ VPP_INPUT_LIBS = \
 
 VPP_INPUT_CFLAGS = \
 	$(LIBVA_CFLAGS) \
+	-I$(top_srcdir) \
```
xuguangxin: seems no place will use VPP_INPUT_CFLAGS , why we need add this

examples/Makefile.am
wudping: thanks @xuguangxin , I will modify it.

examples/Makefile.am
wudping: thanks @xuguangxin , I will modify it.

PR #47, opened 2016-10-10
use __ENABLE_TESTS_GLES to on/off egl compilation

LindaYu17:
Signed-off-by: Linda Yu linda.yu@intel.com

### DIFF
--- a/tests/egl/gles2_help.c
+++ b/tests/egl/gles2_help.c
@@ -353,4 +353,3 @@ void eglRelease(EGLContextType *context)
     eglTerminate(context->eglContext.display);
     free(context);
 }
-
--- a/tests/v4l2decode.cpp
+++ b/tests/v4l2decode.cpp
@@ -203,9 +203,11 @@ static std::vector<struct RawFrameData> rawOutputFrames;
 
 static FILE* outfp = NULL;
 
-#if (!defined(ANDROID) && !defined(__ENABLE_WAYLAND__))
+#if __ENABLE_X11__
 static Display * x11Display = NULL;
 static Window x11Window = 0;
+#endif
+#if __ENABLE_TESTS_GLES__
 static EGLContextType *eglContext = NULL;
 static std::vector<EGLImageKHR> eglImages;
 static std::vector<GLuint> textureIds;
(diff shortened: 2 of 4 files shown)


tests/egl/gles2_help.c
```diff
@@ -23,6 +23,8 @@
 #include "config.h"
 #endif
 
+#if __ENABLE_TESTS_GLES__
+
```
chivakker: This check should be moved out from this file in the same way it was removed from the libyami patches.

LindaYu17:
thanks for the comments, already updated.

tests/v4l2decode.cpp
```diff
@@ -519,9 +521,11 @@ bool takeOneOutputFrame(int fd, int index = -1/* if index is not -1, simple enqu
 #elif __ENABLE_WAYLAND__
         ret = displayOneVideoFrameWayland(fd, &buf);
 #else
-        if (IS_DMA_BUF() || IS_DRM_NAME())
+        if (IS_DMA_BUF() || IS_DRM_NAME()) {
+#ifdef __ENABLE_TESTS_GLES__
             ret = displayOneVideoFrameEGL(fd, buf.index);
-        else
+#endif
+        } else
```
chivakker: when __ENABLE_TESTS_GLES__ is not defined the whole if{} else will be a no operation.  The #ifdef has to be located outside the if (IS_DMA_BUF()...)

tests/v4l2decode.cpp
```diff
@@ -848,6 +852,7 @@ int main(int argc, char** argv)
             }
         }
     } else if (IS_DMA_BUF() || IS_DRM_NAME()) {
+#if __ENABLE_TESTS_GLES__
```
chivakker: same thing here, the else if has to be placed within the __ENABLE_TEST_GLES__ compilation flag.

LindaYu17:
done

xuguangxin:
good to me, thanks Linda and Daniel. We will merge this after 1.0.1 released.

xuguangxin:
meged as 2e5dc8395fed2e720794aea7750f2f542abfeb14, thanks Daniel and Linda

PR #49, opened 2016-10-21
improve grid application. 

xuguangxin:
it will:
1. add frame rate contorl.
2. support multi display
3. display fps on screen

### DIFF
--- a/tests/vppinputasync.cpp
+++ b/tests/vppinputasync.cpp
@@ -15,22 +15,28 @@
  */
 #include "vppinputasync.h"
 
+static const pthread_t INVALID_THREAD_ID = ((pthread_t)-1);
+
 VppInputAsync::VppInputAsync()
-    :m_cond(m_lock), m_eos(false),
-     m_queueSize(0),m_quit(false)
+    : m_cond(m_lock)
+    , m_eos(false)
+    , m_queueSize(0)
+    , m_startupSize(0)
+    , m_thread(INVALID_THREAD_ID)
+    , m_quit(false)
 {
 }
 
 
 SharedPtr<VppInput>
-VppInputAsync::create(const SharedPtr<VppInput>& input, uint32_t queueSize)
+VppInputAsync::create(const SharedPtr<VppInput>& input, uint32_t queueSize, uint32_t startupSize)
 {
     SharedPtr<VppInput> ret;
 
     if (!input || !queueSize)
         return ret;
     SharedPtr<VppInputAsync> async(new VppInputAsync());
-    if (!async->init(input, queueSize)) {
+    if (!async->init(input, queueSize, startupSize)) {
         ERROR("init VppInputAsync failed");
         return ret;
     }
@@ -69,18 +75,47 @@ void VppInputAsync::loop()
    }
 }
 
-bool VppInputAsync::init(const SharedPtr<VppInput>& input, uint32_t queueSize)
+bool VppInputAsync::preloadFrames(uint32_t startupSize)
 {
-    m_input = input;
-    m_queueSize = queueSize;
-    if (pthread_create(&m_thread, NULL, start, this)) {
-        ERROR("create thread failed");
-        return false;
+    uint32_t size;
+    for (size = 0; size < startupSize; size++) {
+        SharedPtr<VideoFrame> frame;
+        if (!m_input->read(frame))
+            break;
+         m_queue.push_back(frame);
+    }
+    if (!size) {
+        ERROR("can't read file from input");
+    }
+    //if the input file do not have enough frames
+    //we will duplicate frames.
+    if (startupSize > size) {
+        uint32_t left = startupSize - size;
+        for (uint32_t i = 0; i < left; i++) {
+            //loop entire queue;
+            uint32_t idx = i % size;
+            m_queue.push_back(m_queue[idx]);
+        }
     }
     return true;
 
 }
 
+bool VppInputAsync::init(const SharedPtr<VppInput>& input, uint32_t queueSize, uint32_t startupSize)
+{
+    m_input = input;
+    m_queueSize = queueSize;
+    m_startupSize = startupSize;
+    if (startupSize) {
+        return preloadFrames(startupSize);
+    } else {
+        if (pthread_create(&m_thread, NULL, start, this)) {
+            ERROR("create thread failed");
+            return false;
+        }
+    }
+    return true;
+}
 bool VppInputAsync::read(SharedPtr<VideoFrame>& frame)
 {
     AutoLock lock(m_lock);
@@ -91,18 +126,24 @@ bool VppInputAsync::read(SharedPtr<VideoFrame>& frame)
     }
     frame = m_queue.front();
     m_queue.pop_front();
+    if (m_startupSize) {
+        //we make a buffer loop here.
+        m_queue.push_back(frame);
+    }
     m_cond.signal();
     return true;
 }
 
 VppInputAsync::~VppInputAsync()
 {
-    {
-        AutoLock lock(m_lock);
-        m_quit = true;
-        m_cond.signal();
+    if (m_thread != INVALID_THREAD_ID) {
+        {
+            AutoLock lock(m_lock);
+            m_quit = true;
+            m_cond.signal();
+        }
+        pthread_join(m_thread, NULL);
     }
-    pthread_join(m_thread, NULL);
 }
 
 bool VppInputAsync::init(const char* inputFileName, uint32_t fourcc, int width, int height)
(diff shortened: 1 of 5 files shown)


xuguangxin:
this is resubmit for https://github.com/01org/libyami/pull/338

xuguangxin:
@YuJiankang , Please help to review this

YuJiankang:
lgtm, +1

examples/grid.cpp
```diff
@@ -319,57 +348,79 @@ bool DrmRenderer::Flipper::init()
...
+                neverSleep = false;
+            } else {
+                if (neverSleep) {
+                    late = !timercmp(&current, &m_nextTime, <);
+                }
+                break;
+            }
+
+        } while (1);
+            /*
```
chivakker: Are planning to remove this commented code?

examples/grid.cpp
```diff
@@ -31,8 +31,6 @@
 #include <Yami.h>
```
chivakker: Maybe you can separate this change from the patch set and get it merged individually.

examples/grid.cpp
```diff
@@ -469,21 +501,126 @@ void DrmRenderer::Flipper::loop()
...
+
         m_lock.acquire();
         if (late) {
             ERROR("late m_fronts.size = %d", (int)m_fronts.size());
         }
         m_pending = true;
         if (m_fps)
         {
-          Timer t("flip");
+          //Timer t("flip");
```
chivakker: remove unneeded commented lines.

examples/grid.cpp
```diff
@@ -469,21 +501,126 @@ void DrmRenderer::Flipper::loop()
...
+            m_start = end;
+        }
+    }
+
+}
+
+int toDec(double fps, double rate)
+{
+    int dec = (int)(fps/rate);
+   // printf("(%f, %f), %d\n", fps, rate, dec);
```
chivakker: remove commented lines.

examples/grid.cpp
```diff
@@ -469,21 +501,126 @@ void DrmRenderer::Flipper::loop()
...
+#define FONT_HEIGHT  48
+void writeDigital(uint32_t* dest, uint32_t pitch, const uint32_t *font )
+{
+    for (int i = 0; i < FONT_HEIGHT; i++) {
+        for (int j = 0; j < FONT_WIDTH; j++) {
+            if (font[j])
+                dest[j] = font[j] | 0xff000000;
+            else
+                dest[j] = 0;
+           //printf("%d, ", (bool)dest[j]);
```
chivakker: ditto.

examples/font.h.LICENSE
```diff
@@ -0,0 +1,202 @@
+
```
chivakker: You don't need to copy all the license.  check other files and do the same.

examples/grid.cpp
```diff
@@ -305,42 +397,94 @@ DrmRenderer::Flipper::~Flipper()
...
+                        timersub(&current, &m_nextTime, &diff);
+                        ERROR("late %d ms",(int)diff.tv_usec/1000);
+			}*/
+                }
+                break;
+            }
+
+        } while (1);
+        /*
+            if (timercmp(&current, &m_nextTime, <)) {
```
chivakker: Remove this commented out code

examples/grid.cpp
```diff
@@ -305,42 +397,94 @@ DrmRenderer::Flipper::~Flipper()
...
+        do {
+            gettimeofday(&current, NULL);
+            timeval refresh;
+            refresh.tv_sec = 0;
+            refresh.tv_usec = 1000 * 1000 / 60;
+            timeval nextRefresh;
+            timeradd(&current, &refresh, &nextRefresh);
+
+            if (timercmp(&nextRefresh, &m_nextTime, < )) {
+                usleep(refresh.tv_usec);
```
chivakker: Can we have instead a clock that is freely running and do the time comparison, then discard frames if late or keep them displayed if no other frame is available.  This  usleep doesn't look ok

examples/grid.cpp
```diff
@@ -305,42 +397,94 @@ DrmRenderer::Flipper::~Flipper()
...
+                neverSleep = false;
+            }
+            else {
+                if (neverSleep) {
+                    late = !timercmp(&current, &m_nextTime, < );
+                    /*       if (late) {
```
chivakker: remove commented code.

examples/grid.cpp
xuguangxin: Hardware decoder is fast, when we are here, it usually early than it should be, so we sleep here to wait render time.  We want know the render fps so we cannot drop frame. In addition, we do not need keep one frame display again since it already a front buffer on fb

examples/grid.cpp
xuguangxin: Better not. This will help debug the performance, but we do not want it printout or even execute in normal case. Therefore, I comment out them here.

examples/grid.cpp
xuguangxin: Any benefit for this?

examples/grid.cpp
xuguangxin: No ,not yet. We do not know which one is better.

examples/font.h.LICENSE
xuguangxin: Ok, I can remove this

examples/grid.cpp
xuguangxin: No, with this line, I will know the flip time when fps assigned. This line have functional. Not a forget remove one

examples/grid.cpp
xuguangxin: this will help debug the digital, better not remove it

xuguangxin:
@chivakker , updated.

xuguangxin:
Hi @chivakker , updated please help to review.
thanks

MicheleL-Intel:
What is the current status?  Do these need to be rebased for closure?

chivakker:
@MicheleL-Intel can you review the patch?

tests/vppinputasync.cpp
```diff
@@ -69,18 +75,47 @@ void VppInputAsync::loop()
...
-        return false;
+    uint32_t size;
+    for (size = 0; size < startupSize; size++) {
+        SharedPtr<VideoFrame> frame;
+        if (!m_input->read(frame))
+            break;
+         m_queue.push_back(frame);
+    }
+    if (!size) {
+        ERROR("can't read file from input");
```
MicheleL-Intel: If the input frame cannot be read, should the function "return false" at this point so that it will not fall thru to the next 'if' statement; unless you intend for the function to fill with duplicate frames regardless of pass/fail.

tests/vppinputasync.h
```diff
@@ -28,18 +28,20 @@ class VppInputAsync : public VppInput
 
     bool read(SharedPtr<VideoFrame>& frame);
 
+    /*startupSize means how many buffers will preload to memory in startup, it will overide queueSize*/
     static SharedPtr<VppInput>
-    create(const SharedPtr<VppInput>& input, uint32_t queueSize);
+    create(const SharedPtr<VppInput>& input, uint32_t queueSize, uint32_t startupSize = 0);
```
MicheleL-Intel: Is it possible to make m_startupSize public to allow direct init/read/modifiy of m_startupSize since the value never changes once it's initialized.  This will eliminate the need for introducing startupSize in every function.

Having two separate variables for the same purpose whereby the values never change requires additional maintenance.

tests/vppinputasync.cpp
```diff
@@ -15,22 +15,28 @@
...
 {
     SharedPtr<VppInput> ret;
 
     if (!input || !queueSize)
         return ret;
     SharedPtr<VppInputAsync> async(new VppInputAsync());
-    if (!async->init(input, queueSize)) {
+    if (!async->init(input, queueSize, startupSize)) {
         ERROR("init VppInputAsync failed");
         return ret;
```
xuguangxin: Hi Michele
the return value of aysnc->init is bool.  and the ret is SharedPtr<VppInput>, we can't assign a bool to SharedPtr<VppInput>.

tests/vppinputasync.cpp
xuguangxin: you are right, if we donot return false, it will crash since m_queue[0] is not validate.I will change this

tests/vppinputasync.h
xuguangxin: No, this is intent to, we can't change m_startupSize at running time. It needs reallocate resource and hold some lock.

tests/vppinputasync.cpp
MicheleL-Intel: I see, that won't work due to mismatched types.  I suppose you can instead set ret=false to handle the failure.

uartie:

PR #50, opened 2016-10-25
Add transcode Latency and bit rate log.

yizhouwei:
add transcode latency for yamitranscode.
enabled by command line option -L <latencyLogFile>
add transcode bitrate
enabled by command line option -R <BitrateLogFile>

### DIFF
--- a/tests/vppinputdecode.cpp
+++ b/tests/vppinputdecode.cpp
@@ -28,6 +28,11 @@ bool VppInputDecode::init(const char* inputFileName, uint32_t /*fourcc*/, int /*
     return true;
 }
 
+void VppInputDecode::setenableLatencyPrintFlag(bool enableFrameLatencyprint)
+{
+    m_enableprintFrameLatency = enableFrameLatencyprint;
+}
+
 bool VppInputDecode::config(NativeDisplay& nativeDisplay)
 {
     m_decoder->setNativeDisplay(&nativeDisplay);
@@ -51,6 +56,15 @@ bool VppInputDecode::config(NativeDisplay& nativeDisplay)
     return status == DECODE_SUCCESS;
 }
 
+void VppInputDecode::getTimeStamp(VideoDecodeBuffer& inputBuffer){
+    if (m_enableprintFrameLatency){
+        struct timeval start;
+        gettimeofday(&start, NULL);
+        int64_t now = 1000000 * start.tv_sec + start.tv_usec;
+        inputBuffer.timeStamp = now;
+    }
+}
+
 bool VppInputDecode::read(SharedPtr<VideoFrame>& frame)
 {
     if (m_first) {
@@ -68,6 +82,7 @@ bool VppInputDecode::read(SharedPtr<VideoFrame>& frame)
         VideoDecodeBuffer inputBuffer;
         Decode_Status status = DECODE_FAIL;
         if (m_input->getNextDecodeUnit(inputBuffer)) {
+            getTimeStamp(inputBuffer);
             status = m_decoder->decode(&inputBuffer);
             if (DECODE_FORMAT_CHANGE == status) {
 
--- a/tests/vppinputoutput.h
+++ b/tests/vppinputoutput.h
@@ -23,6 +23,7 @@
 #include <Yami.h>
 
 #include <stdio.h>
+#include <sys/time.h>
 #include <va/va.h>
 #ifndef ANDROID
 #include <va/va_drm.h>
@@ -209,6 +210,8 @@ public:
                                        const char* codecName = NULL);
     bool getFormat(uint32_t& fourcc, int& width, int& height);
     virtual bool output(const SharedPtr<VideoFrame>& frame) = 0;
+    virtual void printFrameLatency(){};
+    virtual void printBitRate(){};
     VppOutput();
     virtual ~VppOutput(){}
 protected:
@@ -224,7 +227,6 @@ class VppOutputFile : public VppOutput
 {
 public:
     bool config(const SharedPtr<FrameWriter>& writer);
-
     //inherit VppOutput
     bool output(const SharedPtr<VideoFrame>& frame);
     virtual ~VppOutputFile();
--- a/tests/vppoutputencode.cpp
+++ b/tests/vppoutputencode.cpp
@@ -34,6 +34,10 @@ EncodeParams::EncodeParams()
     , numRefFrames(1)
     , idrInterval(0)
     , codec("")
+    , enableprintFrameLatency(false)
+    , FrameLatencyLogFile("Latencylog")
+    , enableprintBitRate(false)
+    , BitRateLogFile("BitRatelog")
     , enableCabac(true)
     , enableDct8x8(false)
     , enableDeblockFilter(true)
@@ -168,7 +172,8 @@ bool VppOutputEncode::config(NativeDisplay& nativeDisplay, const EncodeParams* e
     m_encoder->setNativeDisplay(&nativeDisplay);
     m_mime = m_output->getMimeType();
     setEncodeParam(m_encoder, m_width, m_height, encParam, m_mime, m_fourcc);
-
+    m_bitRate.init(encParam);
+    m_frameLatency.init(encParam);
     Encode_Status status = m_encoder->start();
     assert(status == ENCODE_SUCCESS);
     initOuputBuffer();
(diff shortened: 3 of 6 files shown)


tests/vppinputdecode.cpp
```diff
@@ -51,6 +51,14 @@ bool VppInputDecode::config(NativeDisplay& nativeDisplay)
     return status == DECODE_SUCCESS;
 }
 
+void getTimeStamp(SharedPtr<VideoFrame>& frame, VideoDecodeBuffer inputBuffer){
```
xuguangxin: Better make this as member function and return the time stamp.
because 
```
It may use member variable in future such as frame count.
according to file name it only change time stamp, so it's good to make timestamp as return value
```

tests/vppoutputencode.cpp
```diff
@@ -167,6 +167,15 @@ bool VppOutputEncode::config(NativeDisplay& nativeDisplay, const EncodeParams* e
     return true;
 }
 
+void getFrameDelay(VideoEncOutputBuffer m_outputBuffer){
```
xuguangxin: printFrameDelay maybe better

tests/yamitranscode.cpp
```diff
@@ -342,6 +342,11 @@ class TranscodeTest
         FpsCalc fps;
         uint32_t count = 0;
         while (m_input->read(src)) {
+            struct timeval start;
```
xuguangxin: why we need this since we have timestamp in decode input buffer

tests/vppoutputencode.cpp
```diff
@@ -184,6 +193,7 @@ bool VppOutputEncode::output(const SharedPtr<VideoFrame>& frame)
     }
     do {
         status = m_encoder->getOutput(&m_outputBuffer, drain);
+        getFrameDelay(m_outputBuffer);
```
lizhong1008: Need change to 
if (status == ENCODE_SUCCESS)
getFrameDelay(m_outputBuffer);
That's the reason why you may get same timeStamp multiple time.

tests/vppinputdecode.cpp
xuguangxin: I look at this again. simply remove first argument is ok, but you still need make this as member function.

tests/vppoutputencode.cpp
```diff
@@ -34,6 +34,7 @@ EncodeParams::EncodeParams()
     , numRefFrames(1)
     , idrInterval(0)
     , codec("")
+    , enableFrameLatency(0)
```
xuguangxin: better name it as printLatency or log Latency. enableFrameLatency maybe misleading

tests/vppoutputencode.cpp
```diff
@@ -65,7 +66,8 @@ bool VppOutputEncode::init(const char* outputFileName, uint32_t /*fourcc*/,
     if(!width || !height)
         if (!guessResolution(outputFileName, width, height))
             return false;
-
+    FrameCount = 0;
+    enableLatency = 0;
```
xuguangxin: use true false for bool

tests/vppoutputencode.cpp
```diff
@@ -160,13 +162,37 @@ bool VppOutputEncode::config(NativeDisplay& nativeDisplay, const EncodeParams* e
...
+    int64_t now = 1000000 * end.tv_sec + end.tv_usec;
+    int64_t delay = now - m_outputBuffer.timeStamp;
+    FrameCount++;
+    if (FrameCount == 1){
+        MaxLatency = delay;
+        MinLatency = delay;
+        AveLatency = delay;
+    }
+    else{
+        if (delay > MaxLatency)
```
xuguangxin: better use std::max

tests/vppoutputencode.cpp
```diff
@@ -160,13 +162,37 @@ bool VppOutputEncode::config(NativeDisplay& nativeDisplay, const EncodeParams* e
...
+    FrameCount++;
+    if (FrameCount == 1){
+        MaxLatency = delay;
+        MinLatency = delay;
+        AveLatency = delay;
+    }
+    else{
+        if (delay > MaxLatency) 
+            MaxLatency = delay;
+        if (delay < MinLatency)
```
xuguangxin: better use std::min

tests/vppoutputencode.cpp
```diff
@@ -160,13 +162,37 @@ bool VppOutputEncode::config(NativeDisplay& nativeDisplay, const EncodeParams* e
...
+            MinLatency = delay;
+        AveLatency = (AveLatency * (FrameCount - 1) + delay) / FrameCount;
```
xuguangxin: This may have measurement error and it will do multiple every time. Could we have a total latency then use it to / FrameCount,

tests/vppoutputencode.cpp
```diff
@@ -184,6 +210,9 @@ bool VppOutputEncode::output(const SharedPtr<VideoFrame>& frame)
     }
     do {
         status = m_encoder->getOutput(&m_outputBuffer, drain);
+        if (status == ENCODE_SUCCESS && enableLatency){
```
xuguangxin: Could we merge those two `if (status == ENCODE_SUCCESS )`

tests/vppinputoutput.h
```diff
@@ -280,6 +281,7 @@ class VppOutput
                                        const char* codecName = NULL);
     bool getFormat(uint32_t& fourcc, int& width, int& height);
     virtual bool output(const SharedPtr<VideoFrame>& frame) = 0;
+    virtual void printLatency() = 0;
```
xuguangxin: this not need be pure virtual, in you declear it as `virtual void printLatency(){};` and remove "void printLatency(){};" in VppOutputFile

tests/vppoutputencode.h
```diff
@@ -81,7 +84,14 @@ class VppOutputEncode : public VppOutput
 
 private:
     void initOuputBuffer();
+    void getFrameDelay(VideoEncOutputBuffer m_outputBuffer);
+    
     const char* m_mime;
+    int64_t MaxLatency;
```
xuguangxin: please follow m_ name tradition

tests/vppoutputencode.cpp
yizhouwei: if the frame number is large, int64_t may not hold the total latency

xuguangxin: The latency is diff, it's usually very small, int64_t should enough for all sum

yizhouwei: if we only have a sum but no frame count, we still can't get the average latency

tests/yamitranscode.cpp
```diff
@@ -109,7 +109,7 @@ static bool processCmdLine(int argc, char *argv[], TranscodeParams& para)
         return false;
     }
 
-    while ((opt = getopt_long_only(argc, argv, "W:H:b:f:c:s:i:o:N:h:t:", long_opts,&option_index)) != -1)
+    while ((opt = getopt_long_only(argc, argv, "W:H:b:f:c:s:i:o:N:h:t:L", long_opts,&option_index)) != -1)
```
lizhong1008: don't forget to print_help for option "L"

tests/vppoutputencode.h
```diff
@@ -41,6 +41,7 @@ class EncodeParams {
     int32_t numRefFrames;
     int32_t idrInterval;
     string codec;
+    bool printFrameLatency;
```
lizhong1008: I prefer to change it to enableFrameLatencyPrint.

tests/vppoutputencode.cpp
```diff
@@ -160,13 +162,37 @@ bool VppOutputEncode::config(NativeDisplay& nativeDisplay, const EncodeParams* e
...
+void VppOutputEncode::getFrameDelay(VideoEncOutputBuffer m_outputBuffer){
+    struct timeval end;
+    gettimeofday(&end,NULL);
+    int64_t now = 1000000 * end.tv_sec + end.tv_usec;
+    int64_t delay = now - m_outputBuffer.timeStamp;
+    m_FrameCount++;
+    if (m_FrameCount == 1){
+        m_MaxLatency = delay;
+        m_MinLatency = delay;
+        m_AveLatency = delay;
```
lizhong1008: only one line is needed: m_MaxLatency = m_MinLatency = m_AveLatency = delay;

tests/vppoutputencode.cpp
```diff
@@ -34,6 +34,7 @@ EncodeParams::EncodeParams()
     , numRefFrames(1)
     , idrInterval(0)
     , codec("")
+    , printFrameLatency(0)
```
lizhong1008: it is a bool value, would better init it to printFrameLatency(false)

tests/vppoutputencode.cpp
```diff
@@ -160,13 +162,37 @@ bool VppOutputEncode::config(NativeDisplay& nativeDisplay, const EncodeParams* e
...
+        if (delay > m_MaxLatency) 
+            m_MaxLatency = delay;
+        if (delay < m_MinLatency)
+            m_MinLatency = delay;
+        m_AveLatency = (m_AveLatency * (m_FrameCount - 1) + delay) / m_FrameCount;
+    }
+}
+
+void VppOutputEncode::printFrameLatency(){
+    printf("Frame Count:%ld, MaxLatency:%ldus, MinLatency:%ldus, Average Lantency:%ldus\n", m_FrameCount, m_MaxLatency, m_MinLatency, m_AveLatency);
```
lizhong1008: typo for "Average Lantency", should be Average Latency

lizhong1008: @xuguangxin 1. do we need to print latency of every frame? 2. do we need print it to a file instead of screen?

tests/vppoutputencode.cpp
```diff
@@ -160,13 +162,37 @@ bool VppOutputEncode::config(NativeDisplay& nativeDisplay, const EncodeParams* e
...
+        m_MaxLatency = delay;
+        m_MinLatency = delay;
+        m_AveLatency = delay;
+    }
+    else{
+        if (delay > m_MaxLatency) 
+            m_MaxLatency = delay;
+        if (delay < m_MinLatency)
+            m_MinLatency = delay;
+        m_AveLatency = (m_AveLatency * (m_FrameCount - 1) + delay) / m_FrameCount;
```
xuguangxin: Please do not use m_AveLatency, just sum it up, the latency is small. it will not overflow

tests/yamitranscode.cpp
QuPengfei: some comments in my side:
1.it is better to print out which frame got the max latecny?
2. is it OK to add the log option to print the latency per frame into the log file?

tests/vppoutputencode.h
```diff
@@ -81,7 +88,20 @@ class VppOutputEncode : public VppOutput
 
 private:
     void initOuputBuffer();
+    void getFrameDelay(VideoEncOutputBuffer m_outputBuffer);
+    void getBitRate(VideoEncOutputBuffer m_outputBuffer);
     const char* m_mime;
+    int64_t m_MaxLatency;
```
xuguangxin: This expose too much details to VppOutputEncode , could we use other class like EncoderLantency to hide all details in class? you can refer to
```
class FpsCalc
{
public:
    FpsCalc();
    void addFrame();
    void log();
private:
    int m_frames;
    uint64_t m_start;
    uint64_t m_netStart;
    static const int NET_FPS_START = 5;
};
```
all private member are details, caller do not need know this

tests/vppoutputencode.h
```diff
@@ -81,7 +88,20 @@ class VppOutputEncode : public VppOutput
...
+    int64_t m_MinLatency;
+    int64_t m_TotalLatency;
+    int64_t m_FrameCount;
+    int64_t m_MaxLatencyFrame;
+    int64_t m_totaldataSize;
+    int64_t m_fps;
+    bool enableprintBitRate;
+    bool enableprintLatency;
+    FILE *LatencyLogFile;
```
xuguangxin: please follow our name tradition

tests/yamitranscode.cpp
```diff
@@ -370,7 +382,12 @@ class TranscodeTest
         m_output->output(src);
 
         fps.log();
-
+        if (m_cmdParam.m_encParams.enableprintFrameLatency){
```
xuguangxin: Could we just handle this VppOutputEncoder's deconstructor ? YamiTranscode do not need know this details

tests/vppinputdecode.cpp
```diff
@@ -51,6 +51,13 @@ bool VppInputDecode::config(NativeDisplay& nativeDisplay)
     return status == DECODE_SUCCESS;
 }
 
+void VppInputDecode::getTimeStamp(VideoDecodeBuffer& inputBuffer){
```
xuguangxin: This will break normal decode logical, we should only set this we need measure latency.

yizhouwei: modified

MicheleL-Intel:
What is the current status?

uartie:
Please rebase to fix build

uartie:
Closing due to inactivity and no response by submitter and broken build (needs rebase). Feel free to reopen this pull-request when requested changes and issues are addressed.

PR #57, opened 2016-11-09
yamivpp: color balance setting

wudping:
add setting for hue, saturation, brightness and contrast.

### DIFF
--- a/tests/vpp.cpp
+++ b/tests/vpp.cpp
@@ -95,6 +95,10 @@ public:
         : m_sharpening(SHARPENING_LEVEL_NONE)
         , m_denoise(DENOISE_LEVEL_NONE)
         , m_deinterlaceMode(NULL)
+        , m_hue(COLORBALANCE_LEVEL_NONE)
+        , m_saturation(COLORBALANCE_LEVEL_NONE)
+        , m_brightness(COLORBALANCE_LEVEL_NONE)
+        , m_contrast(COLORBALANCE_LEVEL_NONE)
 #endif
     {
     }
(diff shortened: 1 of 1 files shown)


wudping:
corresponding libyami-utils:
[https://github.com/01org/libyami-utils/pull/57](libyami-utils)

tests/vpp.cpp
```diff
@@ -29,6 +29,11 @@
 
 using namespace YamiMediaCodec;
 
+#define HUE_DEFAULT 50
```
xuguangxin: why we use this as default value, is it special? what if I just want set hue to 50?
Could we just use COLORBALANCE_NONE etc.

tests/vpp.cpp
```diff
@@ -179,6 +192,58 @@ class VppTest
                 case 3:
                     m_deinterlaceMode = optarg;
                     break;
+                case 4:
+                    m_hue = atoi(optarg);
+                    if (m_hue > COLORBALANCE_LEVEL_MAX) {
```
xuguangxin: this check is not needed, we should let the library check this

tests/vpp.cpp
```diff
@@ -246,6 +311,58 @@ class VppTest
                 return false;
             }
         }
+
+        if (m_hue != HUE_DEFAULT) {
+            VPPColorBalanceParameters hue;
```
xuguangxin: so many duplicated code, could we use a function for this?

tests/vpp.cpp
```diff
@@ -29,6 +29,8 @@
 
 using namespace YamiMediaCodec;
 
+#define COLOR_BALANCE_DEFAULT -1
```
xuguangxin: Most ok to me.But this should be COLORBALANCE_LEVEL_NONE. currently it's same value -1, but in future we may change COLORBALANCE_LEVEL_NONE to other value, so use COLORBALANCE_LEVEL_NONE for default is better

wudping: okay， thanks @xuguangxin .
#define COLOR_BALANCE_DEFAULT COLORBALANCE_LEVEL_NONE 
according to your suggestion, I hope the above piece code can work.

xuguangxin:
verify +1

xuguangxin:
merged as 6f70d370e2c9605d1592762ad8c238c4e74514ac

PR #58, opened 2016-11-16
call pooledvideoframeallocator in libyami

LindaYu17:
Signed-off-by: Linda Yu <linda.yu@intel.com>

### DIFF
--- a/tests/vppinputoutput.h
+++ b/tests/vppinputoutput.h
@@ -19,7 +19,7 @@
 #include "common/log.h"
 #include "common/utils.h"
 #include "common/VaapiUtils.h"
-#include "common/videopool.h"
+#include "common/PooledFrameAllocator.h"
 #include <Yami.h>
 
 #include <stdio.h>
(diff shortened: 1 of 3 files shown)


tests/vppinputoutput.h
```diff
@@ -19,7 +19,7 @@
 #include "common/log.h"
 #include "common/utils.h"
 #include "common/VaapiUtils.h"
-#include "common/videopool.h"
+#include <pooledframeallocator.h>
```
xuguangxin: We'd better not make this as public interface, since we may change this later, so just copy the header to yami-utils and call it.

xuguangxin: You can refer to VaapiUtils.h in yami-utils

LindaYu17:
change done @xuguangxin

xuguangxin:
merged as 715fe8cbd81d2d80017489fbae0f3ae51dec5184

PR #59, opened 2016-11-17
yamitranscode: add an option to support h264 simulcast

lizhong1008:
This patch is to support h264 simulcast encoding,
which means independent bitstreams with different priority.
It require to set priority_id and prefix nal unit for each bitstream.

This patch is relative to https://github.com/01org/libyami/pull/659

### DIFF
--- a/tests/vppoutputencode.h
+++ b/tests/vppoutputencode.h
@@ -50,6 +50,9 @@ public:
     int8_t diffQPIB;// B frame qp minus initQP
     uint32_t temporalLayerNum; // svc-t temporal layer number
     uint32_t priorityId; // h264 priority_id in prefix nal unit
+    //h264 simucast means independent bitstreams with different priority.
+    //It require to set priority_id and prefix nal unit for each bitstream.
+    bool enableSimucast;
     EncodeParamsVP9 m_encParamsVP9;
     uint32_t layerBitRate[4]; // specify each scalable layer bitrate
 };
--- a/tests/yamitranscode.cpp
+++ b/tests/yamitranscode.cpp
@@ -62,6 +62,7 @@ static void print_help(const char* app)
     printf("   --btl1 <svc-t layer 1 bitrate: kbps > optional\n");
     printf("   --btl2 <svc-t layer 2 bitrate: kbps> optional\n");
     printf("   --btl3 <svc-t layer 3 bitrate: kbps> optional\n");
+    printf("   --enable-simulcast <enable h264 simulcast, priorityid also should be set)> optional\n");
     printf("   VP9 encoder specific options:\n");
     printf("   --refmode <VP9 Reference frames mode (default 0 last(previous), "
            "gold/alt (previous key frame) | 1 last (previous) gold (one before "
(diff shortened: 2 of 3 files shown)


tests/yamitranscode.cpp
```diff
@@ -62,6 +62,7 @@ static void print_help(const char* app)
     printf("   --btl1 <svc-t layer 1 bitrate: kbps > optional\n");
     printf("   --btl2 <svc-t layer 2 bitrate: kbps> optional\n");
     printf("   --btl3 <svc-t layer 3 bitrate: kbps> optional\n");
+    printf("   --enable-simulcast <enable h264 simulcast, priorityid also should be set)> optional\n");
```
chivakker: Can a new section be included with options for H.264 SVC?  So that the help messages is actually helpful.

tests/vppoutputencode.h
```diff
@@ -50,6 +50,9 @@ class EncodeParams {
     int8_t diffQPIB;// B frame qp minus initQP
     uint32_t temporalLayerNum; // svc-t temporal layer number
     uint32_t priorityId; // h264 priority_id in prefix nal unit
+    //h264 simucast means independent bitstreams with different priority.
+    //It require to set priority_id and prefix nal unit for each bitstream.
+    bool enableSimucast;
```
chivakker: Can we add an overall explanation on how enableSimulcast, priorityID and temporalLayerNum work together?

tests/yamitranscode.cpp
lizhong1008: @chivakker ，we don't support full SVC, we just support svc-t. If temporalLayerNum > 1,  that's is svc-t , we don't need another option for svc-t, temporalLayerNum is enough, and it is very clear.

lizhong1008:
@chivakker "Do you have plans to enable this on yamiencode as well?", No, we mainly use yamitranscode to do encode, yamiencode will be discarded as Guangxin’s long term plan.

lizhong1008:
That means all yamiencode options should also be supported in yamitranscode, but yamitranscode options no need to be supported in yamiencode. @YuJiankang , maybe we can add enableVDEnc in yamitranscode.

tests/vppoutputencode.h
chivakker: exactly that's what I am asking for please, an overall explanation of how all three work.

xuguangxin:
@lizhong1008 @YuJiankang @chivakker @wang-fei , seems yamiencode introduce some confuse, since yamitranscode can support all format yamiencode supported .  Moreover, its performance is better than yamiencode. Do you think it is ok to remove yamiencode right now?

lizhong1008:
@xuguangxin I agree to delete it but we would better have a simple application just demo how to set up encoder pipeline. Yamitranscode seems a little complex for user to learn how to set up encoder pipeline.

xuguangxin:
yes, a simple encoder is in our original plan, just like https://github.com/01org/libyami-utils/blob/master/examples/simpleplayer.cpp. However, we do not have enough resource for this in near future

chivakker:
I would prefer to keep yamiencode as an encoder example.  I understand that it represents duplicating some work with yamitranscode but I think it is still useful in the current way.

xuguangxin:
Hi @chivakker 
Do you think it's ok to remove it when we have simple encoder application? Do you still expect lizhong add similar command line to yamiencode?

xuguangxin:
Hi @chivakker 
any update on this?

MicheleL-Intel:
What is the current status?

uartie:

PR #64, opened 2016-11-30
Add man pages

sebastinas:
They were written by Xu, Guangxin and Yu, Jiankang for the Debian
package.


xuguangxin:
thanks @sebastinas

PR #65, opened 2016-12-02
common: deal with return value of guessFourcc

wudping:
Because the return value of guessFourcc() is 0, here need to set default fourcc
when guessFourcc() returns 0.


wudping:
this commit is based on libyami pull-request #674, please help review, thanks in advance.

wudping:
close this pull-request, because we don't need this any more.

PR #66, opened 2016-12-02
hevc10bit: implement hevc 10 bit-depth transcoding

wudping:
Because add a new bit-depth transcoding, the original logical
to set fourcc becomes more complex. The principle of transcoding
this patch follows as below:
1, the commandline parameter '-s' is just used to set input yuv
   file's fourcc. Only yuv file, it not for other file such as
   H264, VP8 and so on. the file in 264 or other format, their
   fourcc is depended on the first decoded frame.
2, for the output file's fourcc, it firstly comes from the extra
   file name. For example, output file "dd_1920x1080.I420", its
   fourcc is I420. For other can't guess fourcc by its extra name,
   use the fourcc from Vppinput object. For example, output file
   "dd_1920x1080.yuv", that can't guess the fourcc, it will use the
   input file's fourcc. Or output file "dd_1920x1080.265", it also
   uses the input file's fourcc.

### DIFF
--- a/tests/vppinputoutput.cpp
+++ b/tests/vppinputoutput.cpp
@@ -104,11 +104,12 @@ static bool guessFormat(const char* filename, uint32_t& fourcc, int& width, int&
             return false;
         }
     }
-    if (!fourcc)
+    if (!fourcc) {
         fourcc = guessFourcc(filename);
+        if (!fourcc)
+            fourcc = YAMI_FOURCC_I420;
+    }
 
-    if (!fourcc)
-        fourcc = VA_FOURCC('I', '4', '2', '0');
     return true;
 }
 
@@ -181,7 +182,6 @@ SharedPtr<VppOutput> VppOutput::create(const char* outputFileName,
     return output;
 }
 
-
 bool VppOutput::getFormat(uint32_t& fourcc, int& width, int& height)
 {
     fourcc = m_fourcc;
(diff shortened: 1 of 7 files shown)


wudping:
this patch is based on [pull-request #674](https://github.com/01org/libyami/pull/674).
please help to review, thanks in advance.

tests/vppinputoutput.cpp
```diff
@@ -193,17 +193,29 @@ bool VppOutput::getFormat(uint32_t& fourcc, int& width, int& height)
...
-            ERROR("can't guess format from %s", outputFileName);
+    if (!width || !height) {
+        if (!guessResolution(outputFileName, width, height)) {
+            ERROR("failed to guess file(%s) width and height\n", outputFileName);
             return false;
         }
     }
-    m_fourcc = fourcc;
+
+    outputFourcc = guessFourcc(outputFileName);
```
xuguangxin: if the fourcc is not 0, we need use the fourcc, if foucc is 0, we can guess fourcc from the file name

wudping: thanks @xuguangxin , Okay, I will use format parameter foucc firstly.

wudping:
another [pull-request()](https://github.com/01org/libyami-utils/pull/68) is open with the same function.

PR #68, opened 2016-12-05
Yamitrans hevc10bit

wudping:
Implement 10 bits-depth transcoding.
Command line parameter "-s" is just used to set input file's fourcc in case fourcc of input file can not be guessed via its extra name.

### DIFF
--- a/tests/vppinputasync.h
+++ b/tests/vppinputasync.h
@@ -33,6 +33,9 @@ public:
 
     VppInputAsync();
     virtual ~VppInputAsync();
+    virtual int getWidth() { return m_input->getWidth(); }
+    virtual int getHeight() { return m_input->getHeight(); }
+    virtual uint32_t getFourcc() { return m_input->getFourcc(); }
 
     //do not use this
     bool init(const char* inputFileName, uint32_t fourcc, int width, int height);
--- a/tests/vppinputdecode.cpp
+++ b/tests/vppinputdecode.cpp
@@ -75,6 +75,7 @@ bool VppInputDecode::read(SharedPtr<VideoFrame>& frame)
                 const VideoFormatInfo* info = m_decoder->getFormatInfo();
                 m_width = info->width;
                 m_height = info->height;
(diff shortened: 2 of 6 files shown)


wudping:
this patch implements 10 bit-depth transcode.
pleas help to review. 
thanks.

tests/vppinputasync.cpp
```diff
@@ -109,4 +109,25 @@ bool VppInputAsync::init(const char* inputFileName, uint32_t fourcc, int width,
 {
     assert(0 && "no need for this");
     return false;
+}
+
+int VppInputAsync::getWidth()
+{
+    if (m_input && m_width != m_input->getWidth())
```
xuguangxin: m_input is never null. m_input->getWidth() is enough for this

wudping: thanks @xuguangxin , but  if this function:
`bool VppInputAsync::init(const char* inputFileName, uint32_t fourcc, int width, int height)`
is called, m_input will be null.

xuguangxin: it's never be called, you can check the assert in the init function

wudping: But what I concern is that one day in the future, this VppInputAsync::init() is going to be implemented and not to initiate m_input, m_input->getWidth() will raise up a segment error. And the m_input is not the member we have to initiate, because class  VppInputAsync derives from VppInput too, it can override read() function to implement the features which m_input has.

wudping: thanks @xuguangxin, do you think it makes sense?

wudping:
update according to the comments. please help to review again. 
thanks.

wudping:
updated once more. please help to review again.
thanks.

wudping:
sorry @xuguangxin , I should have modified the fourcc initial value with 0, other than YAMI_FOURCC_NV12.
updated, please help to verify and merge.

xuguangxin:
verified +1

PR #72, opened 2017-01-17
enc: resize output buffer size

wudping:
re-alloc a new bigger output buffer for the encoded picture size is
larger than the default.

to fix issue [#657](https://github.com/01org/libyami/issues/657)

### DIFF
--- a/tests/encode.cpp
+++ b/tests/encode.cpp
@@ -45,6 +45,7 @@ int main(int argc, char** argv)
     VideoEncOutputBuffer outputBuffer;
     int encodeFrameCount = 0;
 
+    memset(&outputBuffer, 0, sizeof(VideoEncOutputBuffer));
     if (!process_cmdline(argc, argv))
         return -1;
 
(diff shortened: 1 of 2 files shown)


wudping:
please help to review.

tests/encode.cpp
```diff
@@ -159,6 +159,15 @@ int main(int argc, char** argv)
                 fwrite(MVBuffer.data, MVBuffer.bufferSize, 1, MVFp);
             }
 #endif
+            if (status == ENCODE_BUFFER_TOO_SMALL) {
+                maxOutSize = (maxOutSize * 3) / 2;
+                free(outputBuffer.data);
```
xuguangxin: Is it ok to move this to createOutputBuffer

wudping: sure, accordingly, how about:
`bool createOutputBuffer(VideoEncOutputBuffer* outputBuffer, int maxOutSize)
{
    if(outputBuffer.data){
        free(outputBuffer.data);
        outputBuffer.data = NULL;
    }
    outputBuffer->data = static_cast<uint8_t*>(malloc(maxOutSize));
    if (!outputBuffer->data)
        return false;
    outputBuffer->bufferSize = maxOutSize;
    outputBuffer->format = OUTPUT_EVERYTHING;
    return true;
} `

xuguangxin: yes, and you do not need set "outputBuffer.data = NULL" since you will allocate the data later. Please check the outputBuffer, please make sure you memset the outputBuffer in main(), else free(outputBuffer.data) will crash your programe.

wudping: okay, thanks @xuguangxin .

wudping:
updated according to the suggestions, please help to review.

PR #85, opened 2017-03-16
README: add coverity label

wudping:
add two label for build and coverity in README.md.

to implement [VIZ-9125](https://jira01.devtools.intel.com/browse/VIZ-9125) based on Yizhou's code.


xuguangxin:
Overview section is ill-formatted. 
It should be like Requirements

wudping:
okay, @xuguangxin, I will modify it.

PR #67, opened 2016-12-05
add --startup-size option to yamitranscode and yamvpp 

xuguangxin:
This option will preload yuv to memory. It will help to measure encode and vpp performance.
Example:
`yamivpp -n 1000 --startup-size 100  1920x1080.nv12 320x240.i420 --skip-dump`
It will preload 100 frames to memory, then loop the 100 frame and stop after 1000 frames processed.

PR #87, opened 2017-03-29
add --startup-size option to yamitranscode and yamvpp

wudping:
same as the #67 .

Because @xuguangxin is busy on other thing, I help to fix the review issues.

### DIFF
--- a/tests/vpp.cpp
+++ b/tests/vpp.cpp
@@ -27,6 +27,7 @@
 #include <stdlib.h>
 #include <getopt.h>
 #include <string.h>
+#include <deque>
 
 using namespace YamiMediaCodec;
 
@@ -108,6 +109,7 @@ public:
 #endif
         , m_count(-1)
         , m_startupSize(0)
+        , m_skipDump(false)
     {
     }
     bool init(int argc, char** argv)
@@ -133,11 +135,26 @@ public:
         return bool(m_allocator);
     }
 
+    void flushGPU()
+    {
+        std::deque<SharedPtr<VideoFrame> > frames;
+        SharedPtr<VideoFrame> f;
+        do {
+            f = m_allocator->alloc();
+            if (!f)
+                break;
+            vaSyncSurface(*m_display, f->surface);
+            frames.push_back(f);
+        } while (true);
+    }
+
     bool run()
     {
 
         SharedPtr<VideoFrame> src, dest;
         YamiStatus  status;
+
+        FpsCalc fps;
         uint32_t count = 0;
         while (m_input->read(src)) {
             dest = m_allocator->alloc();
@@ -146,8 +163,12 @@ public:
                 ERROR("vpp process failed, status = %d", status);
                 return true;
             }
-            m_output->output(dest);
+
+            if (!m_skipDump)
+                m_output->output(dest);
             count++;
+            fps.addFrame();
+
             if (count >= m_count)
                 break;
         }
@@ -155,7 +176,12 @@ public:
         dest.reset();
         m_output->output(dest);
 
-        printf("%d frame processed\n", count);
+        //this will force all gpu operation finish. to get real performance
+        if (m_skipDump) {
+            flushGPU();
+        }
+        fps.log();
+
         return true;
     }
 private:
@@ -173,6 +199,7 @@ private:
             { "con", required_argument, NULL, 0 },
             { "startup-size", required_argument, NULL, 0 },
             { "count", required_argument, NULL, 'n' },
+            { "skip-dump", no_argument, NULL, 0 },
             { NULL, no_argument, NULL, 0 }
         };
         int option_index;
@@ -216,6 +243,9 @@ private:
                     break;
                 case 8:
                     m_startupSize = atoi(optarg);
+                    break;
+                case 10:
+                    m_skipDump = true;
                     break;
                 default:
                     usage();
@@ -333,6 +363,7 @@ private:
     int32_t m_contrast;
     uint32_t m_count;
     uint32_t m_startupSize;
+    bool m_skipDump;
 };
 
 void usage()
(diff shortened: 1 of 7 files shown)


wudping:
please help to review, thanks a lot.

tests/vpp.cpp
```diff
@@ -21,18 +21,22 @@
...
 #include <getopt.h>
 #include <string.h>
+#include <deque>
 
 using namespace YamiMediaCodec;
 
 void usage();
 
-SharedPtr<VppInput> createInput(const char* filename, const SharedPtr<VADisplay>& display)
+SharedPtr<VppInput> createInput(const char* filename, const SharedPtr<VADisplay>& display, uint32_t startupSize)
```
wudping: It should be used to specific the frame number:

> bool VppInputAsync::preloadFrames(uint32_t startupSize)
{
    uint32_t size;
    for (size = 0; size < startupSize; size++) {
        SharedPtr<VideoFrame> frame;
        if (!m_input->read(frame))
            break;
        m_queue.push_back(frame);
    }

So uint32_t may be enough.

tests/vpp.cpp
```diff
@@ -125,27 +135,53 @@ class VppTest
         return bool(m_allocator);
     }
 
+    void flushGPU()
+    {
+        std::deque<SharedPtr<VideoFrame> > frames;
+        SharedPtr<VideoFrame> f;
+        do {
+            f = m_allocator->alloc();
+            if (!f)
```
wudping: Thanks for suggestion @MicheleL-Intel, 
But I don't think we should deal with the fail of alloc(), because:
1, alloc() just require a surface from pool, if the pool is  empty, alloc() return NULL, else a frame.

> SharedPtr<VideoFrame> PooledFrameAllocator::alloc()
{
    return  m_pool->alloc();
}

2, m_allocator surface pool's surface elements are created by setformat(). When setformat() failed to create surface from driver, it return false to up layer functions.

> bool PooledFrameAllocator::setFormat(uint32_t fourcc, int width, int height)
{
    std::vector<VASurfaceID> surfaces;
    surfaces.resize(m_poolsize);
......
    VAStatus status = vaCreateSurfaces(*m_display, VA_RT_FORMAT_YUV420, width, height,
            &surfaces[0], surfaces.size(),
            &attrib, 1);

    if (status != VA_STATUS_SUCCESS) {
        ERROR("create surface failed, %s", vaErrorStr(status));
        return false;
    }

tests/vppinputasync.cpp
```diff
@@ -91,18 +126,24 @@ bool VppInputAsync::read(SharedPtr<VideoFrame>& frame)
...
-        AutoLock lock(m_lock);
-        m_quit = true;
-        m_cond.signal();
+    if (m_thread != INVALID_THREAD_ID) {
+        {
+            AutoLock lock(m_lock);
+            m_quit = true;
+            m_cond.signal();
+        }
+        pthread_join(m_thread, NULL);
```
wudping: Thanks @MicheleL-Intel , 
I think VppInputAsync can't be released util the thread terminate, because the thread function start() invokes the member function loop() and indirectly accesses member variables. They are not static.

tests/vpp.cpp
xuguangxin: yes, agree with dongping, The logical here is get and sync all surface in allocator. This will make sure every gpu command finished. The allocate function in allocator only return null when no surface in the pool

tests/vpp.cpp
```diff
@@ -125,27 +135,53 @@ class VppTest
...
+                break;
         }
         //flush output
         dest.reset();
         m_output->output(dest);
 
-        printf("%d frame processed\n", count);
+        //this will force all gpu operation finish. to get real performance
+        if (m_skipDump) {
+            flushGPU();
```
xuguangxin: flushGPU does not return any value, it's a wait function, wait all gpu command flushed. The only thing will be wrong in flushGPU will be vaSyncSurface, if vaSyncSurface return false, it means system trap to a serious issue, we can do nothing about his

tests/vppinputasync.cpp
xuguangxin: no, we only join thread when we created it. if m_thread == INVALID_THREAD_ID, we did not create the thread.

tests/vpp.cpp
xuguangxin: The startupSize is frame number, char is not enough for it, since we may have more than 1000 frames.

xuguangxin:
Hi @wudping 
Why you close this branch?

wudping:
sorry, @xuguangxin, It's a accident. 
I cleaned my repository and branches with script yesterday.
I will reopen it.

uartie:
Please fix compile issues (rebase?):
```
14:52:19,038 INFO  - ../tests/vppoutputencode.cpp: In function 'void setEncodeParam(const std::shared_ptr<YamiMediaCodec::IVideoEncoder>&, int, int, const EncodeParams*, const char*, uint32_t)':
14:52:19,038 INFO  - ../tests/vppoutputencode.cpp:124:36: error: 'VideoRateControlParams {aka struct VideoRateControlParams}' has no member named 'layerBitRate'; did you mean 'bitRate'?
14:52:19,039 INFO  - memcpy(encVideoParams.rcParams.layerBitRate, encParam->layerBitRate,
14:52:19,039 INFO  - ^~~~~~~~~~~~
14:52:19,039 INFO  - ../tests/vppoutputencode.cpp:169:27: error: 'VideoParamsAVC {aka struct VideoParamsAVC}' has no member named 'temporalLayerNum'
14:52:19,039 INFO  - encVideoParamsAVC.temporalLayerNum = encParam->temporalLayerNum;
14:52:19,039 INFO  - ^~~~~~~~~~~~~~~~
```

wudping:
updated, Thanks @uartie .

uartie:
@xuguangxin @wudping is this ready to be merged?

uartie:
Closing... No activity or response from submitter or reviewers for over 30 days.

PR #89, opened 2017-04-05
v4l2: add virtual destructor and check NULL point

wudping:
1, define the base class's destructor as virtual.
2, make sure the pointer is not NULL before accessing.

to fix static code scan issue.


wudping:
pull request policy and shared_ptr issues done, please help to review. 
Thanks

Issue #94, opened 2017-05-13
[regression] yamidecode Segmentation Fault since 52fe395427b3


uartie:
yamidecode produces segfault for any input stream since commit:

52fe395427b3 - 2017-05-12 12:04:45 +0800 - wudping - vp8dec: enable svc-t decoding

```
Starting program: /opt/media/install/bin/yamidecode -i /opt/media/src/media_smoke/media_streams/decodebitstreams/JPEG_RefCompare/010309_0800_4680_nshs.jpg -m 0 -f I420
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".

Program received signal SIGSEGV, Segmentation fault.
0x00007ffff5c246ad in ?? () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
(gdb) bt full
#0  0x00007ffff5c246ad in ?? () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
No symbol table info available.
#1  0x00007ffff5c84529 in std::string::assign(std::string const&) () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
No symbol table info available.
#2  0x00000000004056c9 in operator= (__str="./", this=0x7fffffffe070) at /usr/include/c++/4.8/bits/basic_string.h:547
No locals.
#3  processCmdLine (argc=7, argv=<optimized out>, parameters=parameters@entry=0x7fffffffe050) at decodehelp.cpp:144
        option_index = 32767
        long_opts = {{name = 0x414220 "help", has_arg = 0, flag = 0x0, val = 104}, {name = 0x414225 "capi", has_arg = 0, flag = 0x0, val = 0}, {name = 0x41422a "temporal-layer", has_arg = 1, flag = 0x0, val = 0}, {name = 0x0, 
            has_arg = 0, flag = 0x0, val = 0}}
        opt = <optimized out>
        isSetFourcc = true
        outputFile = "./"
        width = -8144
        height = 32767
#4  0x0000000000404d41 in DecodeTest::init (this=this@entry=0x7fffffffe020, argc=<optimized out>, argv=<optimized out>) at decode.cpp:60
No locals.
#5  0x0000000000404180 in main (argc=<optimized out>, argv=<optimized out>) at decode.cpp:105
[...]

PR #95, opened 2017-05-15
yamidecode: avoid using memset to reset "std::string"

wudping:
If std::string was set with 0, it can't be assigned a new
value on Ubuntu Trusty.

to fix issue [94](https://github.com/01org/libyami-utils/issues/94)


wudping:
Please help to review, thanks.

wudping:
Please help to double check it, Thanks.

PR #98, opened 2017-06-10
tests/decodeoutput: initialize m_file pointer to NULL

uartie:
m_file member pointer is not initialized and will get an arbitrary
value.  If that arbitrary value is not 0, then it fools code like
"if(m_file)" into thinking it's ok to use m_file.

Thus, ensure m_file is initialized to NULL in constructor.

Fixes 01org/libyami#743


wudping:
Thanks @uartie for fixing the bug.

wudping:
+1

wudping:
verify + 1

PR #101, opened 2017-09-18
v4l2: handle on fly resolution change.

xuguangxin:
when we got resolution change notification, we need
1. drain output
2. reallocate surfaces.
3. stream off/on the output


xuguangxin:
@LindaYu17 , please help review this

Issue #102, opened 2017-09-25
remove common/v4l2_wrapper.h


uartie:
`common/v4l2_wrapper.h` is provided by libyami as an installed public header.  The one in libyami-utils does not match the one in libyami and should be removed.

PR #103, opened 2017-09-25
remove v4l2_wrapper header

uartie:
The v4l2_wrapper.h header is provided by libyami as an installed
public header.  Use the installed header instead since the local
one in common is already stale.

Fixes: #102


tests/v4l2encode.cpp
```diff
@@ -27,7 +27,7 @@
 
 #include "common/log.h"
 #include "common/utils.h"
-#include "common/v4l2_wrapper.h"
+#include "libyami/v4l2_wrapper.h"
```
xuguangxin: could we use #include <v4l2_wrapper.h>, just like #include <VideoCommonDefs.h>

uartie:
updated... thanks.

PR #105, opened 2017-09-29
jpegd: skip APP payloads while scanning for start code

uartie:
JPEG APP payloads are notorious for having SOI and EOI markers
in them.  Thus skip over their payloads so we don't
erroneously return an invalid decode unit/image.

VIZ-8219

### DIFF
--- a/tests/decodeinput.cpp
+++ b/tests/decodeinput.cpp
@@ -99,6 +99,8 @@ public:
     bool isSyncWord(const uint8_t* buf);
 private:
     int m_countSOI;
+    int m_skipBytes;
+    bool m_needPayloadLength;
 };
 
 DecodeInput::DecodeInput()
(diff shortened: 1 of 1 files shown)


xuguangxin:
@uartie , this will force the user to skip app play load, will this create a problem for the really user?
Is it possible handling the app payload well in libyami?

uartie:
@xuguangxin the APP "payload" only holds application-specific metadata in a JPEG image.  We need to skip it at this level to properly split JPEG/MJPEG bitstreams on "actual" SOI/EOI markers.  The whole image (including APP payload) is still passed into the decoder/parser to be handled.

uartie:
In other words, SOI and EOI bytes in an APP payload are not real JPEG markers.  The problem was that the image/frame splitter in yamidecode was treating them as "real" markers.

uartie:
@xuguangxin any update?

xuguangxin:
hi @uartie , thanks for explaining. It's clear to me know

PR #106, opened 2017-10-08
configure.ac: add --enable-egl to match with libyami

xuguangxin:
libyami uses --enable-egl to enable egl, but yami-utils uses --enable-tests-gles
for historic reason. this commit adds the --enable-egl to match with libyami.

fix https://github.com/01org/libyami/issues/796


uartie:
In general, I'm ok with changing --enable-tests-gles to --enable-egl to match libyami.  But this does not actually solve 01org/libyami#796.  I have added more details to the bug report to [hopefully] explain the issue better.

uartie:
@xuguangxin please update

uartie:
Closing.  This does not fix actual issue

PR #107, opened 2017-10-11
vpx_enc: set time stamp into every frame

wudping:
In yamiencode, add time stamp into every frame of vp8/vp9 stream files.

To fix issue [#800](https://github.com/01org/libyami/issues/800)

Signed-off-by: dongping wu <dongpingx.wu@intel.com>

### DIFF
--- a/tests/encodeinput.cpp
+++ b/tests/encodeinput.cpp
@@ -259,6 +259,12 @@ const char* EncodeOutputHEVC::getMimeType()
     return YAMI_MIME_HEVC;
 }
 
+void setUint64(uint8_t* header, uint64_t value)
+{
+    uint64_t* h = (uint64_t*)header;
+    *h = value;
+}
+
 void setUint32(uint8_t* header, uint32_t value)
 {
     uint32_t* h = (uint32_t*)header;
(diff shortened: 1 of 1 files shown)


wudping:
please help to review, thanks.

tests/encodeinput.cpp
```diff
@@ -294,11 +300,12 @@ bool EncodeOutputVPX::init(const char* outputFileName, int width, int height, in
     return EncodeOutput::write(header, sizeof(header));
 }
 
-bool EncodeOutputVPX::write(void* data, int size)
+bool EncodeOutputVPX::write(void* data, int size, uint64_t timeStamp)
 {
     uint8_t header[12];
     memset(header, 0, sizeof(header));
     setUint32(header, size);
+    setUint64(&header[4], timeStamp);
```
xuguangxin: we can't write the timestamp directly to the frame header. It will generate a wrong bitstream if you use mp4 as input.
The ivf timestamp depends on two things
1. the time scale in the file header.
2. the time stamp in the frame header. 
we write the fps as time scale in the file header, The time stamp need be the frame number.

wudping: Thanks for your details.
Is it okay if  I will replace timeStamp with frame number?

wudping:
updated, please help to review again.

PR #111, opened 2017-11-20
test/yamitranscode: fix decode output file format is always I420.

feiwan1:
Use specific fourcc format or format resolved from output file
to create vpp output in yamitranscode.

Signed-off-by: Wang,Fei <fei.w.wang@intel.com>

### DIFF
--- a/tests/yamitranscode.cpp
+++ b/tests/yamitranscode.cpp
@@ -395,7 +395,7 @@ public:
             return false;
         }
         m_input = createInput(m_cmdParam, m_display);
-        m_output = createOutput(m_cmdParam, m_display, m_input->getFourcc());
+        m_output = createOutput(m_cmdParam, m_display, m_cmdParam.fourcc);
         if (!m_input || !m_output) {
             ERROR("create input or output failed");
             return false;


tests/yamitranscode.cpp
```diff
@@ -395,7 +395,7 @@ class TranscodeTest
             return false;
         }
         m_input = createInput(m_cmdParam, m_display);
-        m_output = createOutput(m_cmdParam, m_display, m_input->getFourcc());
+        m_output = createOutput(m_cmdParam, m_display, m_cmdParam.fourcc);
```
uartie: m_cmdParam.fourcc option is a little confusing... it is applied to VppInput too.  Should input and output have their own fourcc cmdline options?  How would user transcode NV12 input to YV12 output?

uartie:
What use-case are you trying to fix?  Can you give an example?

tests/yamitranscode.cpp
feiwan1: @uartie , yamitranscode can get the output fourcc by the output filename. if without this patch, when I use 'yamitranscode -i 1920x1080.NV12 1920x1080.YV12' want to convert NV12 to YV12. But the output file 1920x1080.YV12 actually is still NV12 format.

uartie:
please rebase to fix CI builds

tests/yamitranscode.cpp
xuguangxin: @feiwan1 , it's not right, we need m_input->getFourcc() to tell output, we need 10 bits encoder.
maybe we can remvoe -s command in yamitranscode, it's not even listed in help.
and we may consider removing VppOutputFile branch in createOutput. 
The transcode output is raw frame does not make sense.

uartie:
Closing as inactive... please resubmit if you feel this is still relevant.

PR #112, opened 2017-11-20
Change FILE* to std::ofstream and std::ifstream objects

uartie:
Fixes 01org/libyami#744

### DIFF
--- a/tests/v4l2encode.cpp
+++ b/tests/v4l2encode.cpp
@@ -17,7 +17,8 @@
 #include "config.h"
 #endif
 
-#include <stdio.h>
+#include <iostream>
+#include <fstream>
 #include <string.h>
 #include <stdlib.h>
 #include <assert.h>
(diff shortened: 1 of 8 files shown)


uartie:
@xuguangxin @YuJiankang @wudping please help to review

wudping:
it's good, for me.

tests/v4l2encode.cpp
```diff
@@ -147,23 +148,23 @@ bool feedOneInputFrame(int fd, int index = -1 /* if index is not -1, simple enqu
...
-    static FILE* fp = NULL;
+    static std::ofstream m_ofs;
 
-    if(!fp) {
-        fp = fopen(outputFileName, "w+");
-        if (!fp) {
+    if(!m_ofs.is_open()) {
+        m_ofs.open((outputFileName, std::ofstream::out | std::ofstream::trunc);
+        if (!m_ofs) {
             fprintf(stderr, "fail to open file: %s\n", outputFileName);
```
wudping: hi, should this line use std::cerr too?

uartie: It's a little out of scope for these changes.  I agree, we should replace all fprintf usage... but that should be handled in another patch series.

PR #113, opened 2017-12-20
Fast boot 1215 working

wudping:
Modularize the libyami-utils code, so that yamidecode, yamiencode, yamivpp, simpleplayer and so on, can be compiled separately. It helps to reduce the size of those each application binary.

### DIFF
--- a/tests/Makefile.am
+++ b/tests/Makefile.am
@@ -1,4 +1,4 @@
-bin_PROGRAMS = yamidecode yamiencode yamivpp yamitranscode
+bin_PROGRAMS = yamidecode yamiencode yamivpp yamitranscode yamiplayer
 if ENABLE_V4L2
 bin_PROGRAMS += v4l2encode v4l2decode
 endif
@@ -28,6 +28,14 @@ YAMI_DECODE_LIBS = \
 	$(YAMI_COMMON_LIBS) \
 	$(NULL)
 
+YAMI_PLAYER_INPUT_SOURCES = \
+	decodeinput.cpp \
+	$(NULL)
+
+YAMI_PLAYER_LIBS = \
+	$(YAMI_COMMON_LIBS) \
+	$(NULL)
+
 if ENABLE_V4L2
 AM_CFLAGS += \
 	$(LIBYAMIV4L2_CFLAGS) \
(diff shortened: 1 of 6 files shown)


wudping:
please help to review, thanks.

wudping:
This pull-request [libyami-pull-911](https://github.com/01org/libyami/pull/819). So there is an compilation error:
09:21:43,558 INFO  - simpleplayer.cpp: In member function 'bool SimplePlayer::run()':
09:21:43,558 INFO  - simpleplayer.cpp:175:26: error: 'struct VideoConfigBuffer' has no member named 'noNeedExtraSurface'
09:21:43,558 INFO  - configBuffer.noNeedExtraSurface = true;

configure.ac
```diff
@@ -174,6 +174,95 @@ AC_ARG_ENABLE(oclblender,
 AM_CONDITIONAL(BUILD_OCL_BLENDER,
     [test "x$enable_oclblender" = "xyes"])
 
+APPS=""
+
+dnl decoder
```
xuguangxin: why we need this, could we just compile all things?

examples/simpleplayer.cpp
```diff
@@ -20,38 +20,150 @@
...
 #include <va/va_x11.h>
+#endif
+#ifdef __ENABLE_WAYLAND__
+#include <va/va_wayland.h>
+#endif
 
 using namespace YamiMediaCodec;
 
+#define CPPPRINT(...) std::cout << __VA_ARGS__ << std::endl
+#define FILE_OUTPUT 0
```
xuguangxin: why we need add file output, could we use yamidecode to do this?

examples/simpleplayer.cpp
```diff
@@ -163,16 +489,123 @@ class SimplePlayer
...
     SharedPtr<IVideoDecoder> m_decoder;
     SharedPtr<DecodeInput> m_input;
     int m_width, m_height;
+    bool m_eos;
+    int m_drmFd;
+    std::ofstream m_ofs;
+    uint64_t m_frameNum;
+    SimplePlayerParameter m_parameters;
+#ifdef __ENABLE_X11__
+    SharedPtr<Display> m_display;
```
xuguangxin: could we have some abstract and isolate all X11/Wayland specific operation in class?

wudping: this code is used in fast-boot, so we try to keep the code as small as possible, and not to define classes.

configure.ac
wudping: because libyami may not compile decoder or encoder, so we have to disable decoder in case that libyami disable all decoders. In fast-boot case, users only need decoder, so we should disable encoder and other functional code except for decoder.

wudping:
hi @xuguangxin , all,
code are managed in classes, please help to review again.

uartie:
Closing as inactive... please resubmit if you feel this is still relevant.

PR #119, opened 2018-02-07
Revert "decode avformat: av_register_all is deprecated since 58.9.100"

uartie:
Reverts intel/libyami-utils#118

Forgot to push local update before
submitting pull-request.


uartie:
CI builds will fail without commit.  I will be submitting an updated patch.

uartie:
nevermind... i'll make change on top of merged commit

PR #121, opened 2018-03-19
Avoid deprecated libtool macro

kwizart:
Signed-off-by: Nicolas Chauvet <kwizart@gmail.com>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -186,8 +186,7 @@ AC_PROG_CPP
 AC_PROG_INSTALL
 AC_PROG_LN_S
 AC_PROG_MAKE_SET
-AC_PROG_LIBTOOL
-#LT_INIT
+LT_INIT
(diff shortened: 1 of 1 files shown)


xuguangxin:
@kwizart , thanks for your patch. could you write down, how to reproduce your issue ?

kwizart:
This is about using a deprecated macro in configure.ac as per the fedora review guidelines, it's not allowed anymore to use it.
https://bugzilla.redhat.com/show_bug.cgi?id=1553377#c1

The autotools documentation:
https://www.gnu.org/software/libtool/manual/html_node/LT_005fINIT.html

There is no current OS still only using the old macro, so it will works everywhere.
(tested on EL6 using autoconf268 from epel)

xuguangxin:
@kwizart thanks for detailed description

PR #122, opened 2018-04-10
travis: fix the build issue since we move from 01org to intel

xuguangxin:


xuguangxin:
@uartie travis passed. this means it will rebase and compile automatically, right? could we do this in internal build server?
thanks

uartie:
Travis CI only passes because it does not compile all the code (e.g. no `--enable-avformat`).  The Intel CI has `--enable-avformat` (and all other configure flags enabled)... this patch is based on an out-of-date tree that is missing patches for avformat fixes.  Thus, compile is broken there.  None of the CI's do automatic rebase of pull-requests.  It is best practice to always ensure pull-requests are based on latest tip.  I agree that the patch in this pull-request is not the cause for the build failure, but the principal of good practice should be followed, nonetheless.

xuguangxin:
thanks for detailed explanation. rebased.

PR #123, opened 2018-04-28
add built in md5 sum calculator. 

xuguangxin:
Android does not have libbsd. so we need use built-in md5 calculator.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -148,10 +148,6 @@ fi
 AM_CONDITIONAL(ENABLE_AVFORMAT,
     [test "x$enable_avformat" = "xyes"])
 
-AC_ARG_ENABLE(md5,
-    [AC_HELP_STRING([--enable-md5], [enable generate md5 by per frame@<:@default=yes@:>@])],
-    [], [enable_md5="yes"])
-
 dnl encoder getmv
 AC_ARG_ENABLE(getmv,
     [AC_HELP_STRING([--enable-getmv],
@@ -247,17 +243,15 @@ if test "$enable_avformat" = "yes"; then
     PKG_CHECK_MODULES(LIBAVFORMAT,  [libavformat libavcodec libavutil])
 fi
 
-#check openssl
-if test "$enable_md5" = "yes"; then
-    PKG_CHECK_MODULES([LIBBSD], [libbsd],
-        [AC_DEFINE([__ENABLE_MD5__], [1],
-            [Defined to 1 if MD5 API and --enable-md5[default] are enabled])],
-        [])
+
+PKG_CHECK_MODULES([LIBBSD], [libbsd], [have_bsd="yes"], [have_bsd="no"])
+AM_CONDITIONAL([HAVE_LIBBSD], [test "x$have_bsd" = xyes])
+if test "$have_bsd" = "yes"; then
+   AC_DEFINE(HAVE_LIBBSD, 1, [Define if libbsd is available])
 fi
 
-PKG_CHECK_MODULES([LIBYAMI], [libyami >= 0.5.2])
 
-AM_CONDITIONAL(ENABLE_MD5, test "x$enable_md5" = "xyes")
+PKG_CHECK_MODULES([LIBYAMI], [libyami >= 0.5.2])
 
 # Checks for library functions.
 AC_FUNC_MALLOC
--- a/tests/Makefile.am
+++ b/tests/Makefile.am
@@ -58,7 +58,7 @@ if ENABLE_EGL
 YAMI_DECODE_LIBS += $(LIBEGL_LIBS) $(LIBGLES2_LIBS)
 endif
 
-if ENABLE_MD5
+if HAVE_LIBBSD
 YAMI_DECODE_LIBS += $(LIBBSD_LIBS)
 endif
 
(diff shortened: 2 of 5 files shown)


uartie:
Perhaps we should add option `--with-md5=[libbsd|builtin]` or `--enable-md5=[libbsd|builtin].  This way we can still have configure fail if user requests bsd solution but forgets to install the dependency.

xuguangxin:
@uartie , thanks, this is a very reasonable suggestion. I will  change it

configure.ac
```diff
@@ -247,13 +247,9 @@ if test "$enable_avformat" = "yes"; then
...
-#check openssl
-if test "$enable_md5" = "yes"; then
-    PKG_CHECK_MODULES([LIBBSD], [libbsd],
-        [AC_DEFINE([__ENABLE_MD5__], [1],
-            [Defined to 1 if MD5 API and --enable-md5[default] are enabled])],
-        [])
-fi
+
+PKG_CHECK_MODULES([BSD], [libbsd], [have_bsd="yes"], [have_bsd="no"])
+AM_CONDITIONAL(HAVE_LIBBSD, test x"$have_bsd" = "xyes")
```
uartie: AM_CONDITIONAL only defines it for makefiles to use.  AC_DEFINE defines it in config.h for sourcecode to use.

It seems we need two defines for sourcecode... one if --enable-md5 is present and one if libbsd is present.

We already have ENABLE_MD5 for makefiles

tests/md5.c
```diff
@@ -0,0 +1,291 @@
...
+ * configuration, and the function prototypes match OpenSSL's.  No code from
+ * Colin Plumb's implementation has been reused; this comment merely compares
+ * the properties of the two independent implementations.
+ *
+ * The primary goals of this implementation are portability and ease of use.
+ * It is meant to be fast, but not as fast as possible.  Some known
+ * optimizations are not included to reduce source code size and avoid
+ * compile-time configuration.
+ */
+
```
uartie: Missing:
```
#ifdef HAVE_CONFIG_H
#include "config.h"
#endif
```

tests/Makefile.am
```diff
@@ -95,7 +95,7 @@ YAMI_VPP_CFLAGS = \
 
 yamidecode_LDADD    = $(YAMI_VPP_LIBS)
 yamidecode_LDFLAGS  = $(YAMI_VPP_LDFLAGS)
-yamidecode_SOURCES  = decode.cpp decodehelp.cpp $(DECODE_INPUT_SOURCES) decodeoutput.cpp vppinputoutput.cpp vppinputdecode.cpp vppoutputencode.cpp encodeinput.cpp encodeInputCamera.cpp encodeInputDecoder.cpp vppinputdecodecapi.cpp
+yamidecode_SOURCES  = decode.cpp decodehelp.cpp $(DECODE_INPUT_SOURCES) decodeoutput.cpp vppinputoutput.cpp vppinputdecode.cpp vppoutputencode.cpp encodeinput.cpp encodeInputCamera.cpp encodeInputDecoder.cpp vppinputdecodecapi.cpp md5.c
```
uartie: We only want to compile md5.c if ENABLE_MD5 is defined from AM_CONDITIONAL in configure.ac

tests/decodeoutput.cpp
```diff
@@ -333,7 +326,6 @@ bool DecodeOutputDump::output(const SharedPtr<VideoFrame>& frame)
     return m_output->output(dest);
 }
 
-#if __ENABLE_MD5__
 class DecodeOutputMD5 : public DecodeOutputFile {
```
uartie: DecodeOutputMD5 needs defined only if --enable-md5

tests/md5.c
```diff
@@ -0,0 +1,291 @@
...
+#if !defined(HAVE_OPENSSL) || !defined(HAVE_LIBBSD)
```
uartie: Where is HAVE_OPENSSL defined?

HAVE_LIBBSD not defined in sourcecode context (see configure.ac comment)

tests/md5.h
```diff
@@ -0,0 +1,63 @@
...
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted.
+ *
+ * There's ABSOLUTELY NO WARRANTY, express or implied.
+ *
+ * See md5.c for more information.
+ */
+
+#ifdef HAVE_OPENSSL
```
uartie: Where is HAVE_OPENSSL defined?

tests/md5.h
```diff
@@ -0,0 +1,63 @@
...
+#include <openssl/md5.h>
+#elif defined(HAVE_LIBBSD)
```
uartie: HAVE_LIBBSD is not defined in sourcecode context (see configure.ac comment)

tests/decodeoutput.cpp
```diff
@@ -22,16 +22,9 @@
...
-// current environment, ignoring attributes and then poping to restore the
-// environment.
-#pragma GCC diagnostic push
-#pragma GCC diagnostic ignored "-Wattributes"
-#include <bsd/md5.h>
-#pragma GCC diagnostic pop
-#endif
+extern "C" {
+#include "md5.h"
+}
```
uartie: need only include it if --enable-md5

tests/decodeoutput.cpp
```diff
@@ -915,11 +905,9 @@ DecodeOutput* DecodeOutput::create(int renderMode, uint32_t fourcc, const char*
 {
     DecodeOutput* output;
     switch (renderMode) {
-#ifdef __ENABLE_MD5__
     case -2:
```
uartie: should only be available if --enable-md5

configure.ac
xuguangxin: I want to change the logical here.
We can autodetect libbsd, if we do not have libbsd we can use built-in md5 sum. 
this logical should work fine in env with or without libbsd.

tests/Makefile.am
xuguangxin: this will introduce complicity, how about let md5 build by default. if we have libbsd we will use libbsd, if we do not have we fallback to a built in md5 sum function(slow one)

tests/md5.c
xuguangxin: yeah, sorry for miss this

tests/md5.h
xuguangxin: no, this just original code. I did not change it.

xuguangxin:
@uartie , thanks for your careful review.  done.
the basic idea behind this change does not let the user choose too many things.
we automatically detect libbsd. If we do not have libbsd we will choose built-in md5 sum.
Force user uses libbsd is not a friendly way.

configure.ac
```diff
@@ -247,17 +247,15 @@ if test "$enable_avformat" = "yes"; then
...
+
+PKG_CHECK_MODULES([LIBBSD], [libbsd], [have_bsd="yes"], [have_bsd="no"])
+AM_CONDITIONAL([HAVE_LIBBSD], [test "x$have_bsd" = xyes])
+if test "$have_bsd" = "yes"; then
+   AC_DEFINE(HAVE_LIBBSD, 1, [Define if libbsd is available])
 fi
 
-PKG_CHECK_MODULES([LIBYAMI], [libyami >= 0.5.2])
 
-AM_CONDITIONAL(ENABLE_MD5, test "x$enable_md5" = "xyes")
```
uartie: Also need to remove `--enable-md5` option at lines 151-153

PR #125, opened 2018-05-15
Fix low latency

xuguangxin:
If we know we got a entire frame, we can pass the flag to decoder to tell them stop search for next frame. this need work with https://github.com/intel/libyami/pull/849


uartie:
Please ignore `intel/ci/fedora/26/pr/compile` and `intel/ci/ubuntu/artful/pr/compile` check results.  These two distros are EOL and have been removed from CI and therefore will not get updated.

PR #135, opened 2018-09-21
Properly initialize VideoDecodeBuffer

### LINKED ISSUE #134: [regression] Arbitrary initialization of VideoDecodeBuffer::flag causes incorrect AVC/HEVC decoding
Arbitrary initialized VideoDecodeBuffer in VppInputDecode::read can result in invalid AVC/HEVC decoding since libyami commit:

```
commit 83f9fa79c856cebf98ca9b0d6dfc0b0acba9f443
Author: XuGuangxin <Guangxin.Xu@intel.com>
Date:   Fri May 4 12:44:00 2018 +0800

    h264, h265: add VIDEO_DECODE_BUFFER_FLAG_FRAME_END to inidicate frame end
    
    unless we got AUD or SPS/PPS in stream, we can't know we have a frame.
    but in up level, we may know we have entire frame. so we add a frame to
    tell decoder about this
```

The arbitrary value can erroneously trigger VIDEO_DECODE_BUFFER_FLAG_FRAME_END in AVC/HEVC decoder, resulting in wrong decode output or decode errors.

uartie:
libyami introduced VIDEO_DECODE_BUFFER_FLAG_FRAME_END and
uses it in avc/hevc decoder to stop frame decoding in
some cases.  Thus, it is necessary to properly initialize
the VideoDecodeBuffer so that VIDEO_DECODE_BUFFER_FLAG_FRAME_END
is not erroneously and arbitrarily set.

Fixes #134

### DIFF
--- a/examples/androidplayer.cpp
+++ b/examples/androidplayer.cpp
@@ -141,6 +141,8 @@ public:
         assert(status == DECODE_SUCCESS);
 
         VideoDecodeBuffer inputBuffer;
+        memset(&inputBuffer, 0, sizeof(inputBuffer));
+
         while (m_input->getNextDecodeUnit(inputBuffer)) {
             status = m_decoder->decode(&inputBuffer);
             if (DECODE_FORMAT_CHANGE == status) {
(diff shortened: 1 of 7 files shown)


uartie:
@wangzj0601 can this be included in 1.3.1 release?

xuguangxin:
U.Artie, +1, thanks

